PR #6, opened 2016-05-18
Allow producer and consumer props to be passed to services as a map instead of a file

lindong28:
Previously we pass producer props to ProduceService as a file. Same thing for ConsumeService as well. However, this approach is not convenient if user wants to instantiate mutliple ProduceService in the same Kafka Monitor JVM process, since it requires users to create multiple files. This patch aims to simplify this process by allowing user to provide producer/consumer properties as a map for each service in the kafka-monitor.properties file that is passed to Kafka Monitor.


lindong28:
@becketqin Do you have time to review this patch?

becketqin:
LGTM.

PR #17, opened 2016-07-23
Improve the way the config value is passed to producer/consumer

lindong28:


lindong28:
Some configs specified for EndToEndTest in the json-formatted properties, such as zookeeper.connect, should be passed to ProduceService/ConsumerService even for KMBaseConsumer class that is implemented outside Kafka Monitor. Otherwise, users would have to specify zookeeper.connect twice in the config.

lindong28:
@becketqin @onurkaraman Can you review the patch?

PR #18, opened 2016-07-26
Measure consume availability

lindong28:


services/src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -196,43 +202,72 @@ public boolean isRunning() {
...
     private final Sensor _recordsDelay;
+    private final Sensor _recordsDelayed;
+
+    public ConsumeMetrics(Metrics metrics, final Map<String, String> tags) {
+      this.metrics = metrics;
 
-    public ConsumeMetrics(Metrics metrics, Map<String, String> tags) {
       _bytesConsumed = metrics.sensor("bytes-consumed");
-      _bytesConsumed.add(new MetricName("bytes-consumed-rate", METRIC_GROUP_NAME, tags), new Rate());
+      _bytesConsumed.add(new MetricName("bytes-consumed-rate", METRIC_GROUP_NAME, "The average number of bytes per second that are consumed", tags), new Rate());
```
becketqin: Should this be using `metrics.metricName()`?

services/src/main/java/com/linkedin/kmf/services/configs/ConsumeServiceConfig.java
```diff
@@ -42,6 +42,10 @@
   public static final String CONSUMER_PROPS_CONFIG = "consume.consumer.props";
   public static final String CONSUMER_PROPS_DOC = "The properties used to config consumer in consume service.";
 
+  public static final String LATENCY_SLA_MS_CONFIG = "consume.latency.sla.ms";
```
becketqin: Can we rename this to CONSUMER_SLA_MS_CONFIG?

services/src/main/java/com/linkedin/kmf/services/ConsumeService.java
lindong28: I have considered this. But `metrics.metricName()` is not available in kafka 0.9.0.1. We can replace all usage of deprecated Kafka APIs the next time we upgrade Kafka Monitor's dependency on Kafka.

services/src/main/java/com/linkedin/kmf/services/configs/ConsumeServiceConfig.java
lindong28: Do you think `consume.latency.sla.ms` is a good config name? Currently the variable name is named after the `consume.` part, e.g. LATENCY_SLA_MS for `.latency.sla.ms`. Because all we already know that variables in `ConsumeServiceConfig.java` are used for ConsumeService.

becketqin: Yeah, the configuration name is a little weird. It may be better to change all the `consume.consumer.*`  to `consumer.*`

lindong28: There is reason for have the prefix `consume` in the config. For example, we have these two configs in `ConsumeServiceConfig` - `consume.latency.percentile.max.ms` and `consume.consumer.props`. And we have these two configs in `ProduceServiceConfig` -  `produce.thread.num` and `produce.thread.num`. The prefix `consume.*` (or `produce.*`) is used to indicate which service this config belongs to. `consume.latency.percentile.max.ms` is directly used for that ConsumeService, while `consume.consumer.props` is used for the consumer in the ConsumeService. It is because of this kind of structure that we have named `consume.consumer.* ` instead of simply `consumer.*`.

becketqin: I see. Thanks for the explanation. I suppose that user will pass in a flat map property file so we need the service name to distinguish the configurations with the same name in different services. Otherwise the patch LGTM.

lindong28: Yes, it is sort of like that. The user can pass in a map which may potentially contain configuration of the same name in difference services. It is useful to put service name in the prefix of config name to distinguish between configs of difference services. See https://github.com/linkedin/kafka-monitor/blob/master/config/kafka-monitor.properties for example config file. In this example, `zookeeper.connect ` is a config used by `BasicEndToEndTest ` while `produce.producer.props ` is a config used by `ProduceService` started as part of `BasicEndToEndTest`.

lindong28: @becketqin Many thanks for taking time to review the patch!

PR #20, opened 2016-09-02
remove external dependencies for js and css , load locally

ajayk:


lindong28:
@ajayk Sorry for late reply. I didn't notice the pull request until now.

It will increases the size of repository if we replace external dependency with local images and javascripts. It also requires us to replace all files if we want to depend on a higher version of javascript in the future. Can you explain the benefits of this approach? Is this because the machine which runs the javascript is not able to fetch the dependent files from network?

ajayk:
@lindong28 the use case was to run the kafka monitor ui without any dependency , 
As you clearly stated that some machines will not have a outbound connections to the internet .

lindong28:
@ajayk Typically we don't want to copy&paste code from dependent library to our own repository for good reasons. It is preferred to download dependent library from remote repository, as we are doing with Java dependency specified in gradle. We should probably follow the same convention here with js dependency. There is trade-off between requirement of network connectivity and easiness of dependency management.

BTW, we provide this simple (quick-start) GUI to demonstrate the metrics that you can read from Kafka Monitor. However, it is not mean to provide production use. For example, you can not read metrics older than 30 seconds in the past unless you modify the index.html manually. Given this background, it is probably better for you to clone the repository and tune the index.html as needed. Does this make sense?

ajayk:
@lindong28 Makes Sense 

I think currently json2.js and jolokia and chart.js dependecy should be referred through CDN as well

lindong28:
@ajayk Yeah, I tried to import them via CDN as well. It is just that I couldn't find a popular/stable public link that provides them.

PR #22, opened 2016-09-30
Automatically create a monitoring topic if one has not already been created.

smccauliff:
This is the first of several changes needed to automatically distribute monitored partitions over the all brokers.  This creates a topic if one has not already been created with twice as many partitions as brokers.

This change has been tested by starting a cluster of three brokers and then verifying that the monitor starts successfully and the log message about the topic starting shows up in the log.

This change should not break any existing processes used to create monitoring topics as the monitor will still use an existing topic for monitoring if it exists.

### DIFF
--- a/config/kafka-monitor.properties
+++ b/config/kafka-monitor.properties
@@ -39,7 +39,7 @@
     "zookeeper.connect": "localhost:2181",
     "bootstrap.servers": "localhost:9092",
     "produce.record.delay.ms": 100,
-    "produce.topic.autoTopicCreationEnabled": false,
+    "produce.topic.autoTopicCreationEnabled": true,
     "produce.topic.autoTopicReplicationFactor" : 1,
     "produce.producer.props": {
       "client.id": "kmf-client-id"
--- a/config/kafka-monitor.properties
+++ b/config/kafka-monitor.properties
@@ -41,7 +41,7 @@
     "produce.record.delay.ms": 100,
     "produce.topic.autoTopicCreationEnabled": true,
     "produce.topic.autoTopicMinIsr" : 2,
-    "produce.topic.autoautoTopicReplicationFactor" : 3,
+    "produce.topic.autoTopicReplicationFactor" : 3,
     "produce.producer.props": {
       "client.id": "kmf-client-id"
     },
--- a/config/kafka-monitor.properties
+++ b/config/kafka-monitor.properties
@@ -40,8 +40,7 @@
     "bootstrap.servers": "localhost:9092",
     "produce.record.delay.ms": 100,
     "produce.topic.autoTopicCreationEnabled": true,
-    "produce.topic.autoTopicMinIsr" : 2,
-    "produce.topic.autoTopicReplicationFactor" : 3,
+    "produce.topic.autoTopicReplicationFactor" : 1,
     "produce.producer.props": {
       "client.id": "kmf-client-id"
     },
--- a/config/kafka-monitor.properties
+++ b/config/kafka-monitor.properties
@@ -39,6 +39,9 @@
     "zookeeper.connect": "localhost:2181",
     "bootstrap.servers": "localhost:9092",
     "produce.record.delay.ms": 100,
+    "produce.topic.autoTopicCreationEnabled": true,
+    "produce.topic.autoTopicMinIsr" : 2,
+    "produce.topic.autoautoTopicReplicationFactor" : 3,
     "produce.producer.props": {
       "client.id": "kmf-client-id"
     },
--- a/config/kafka-monitor.properties
+++ b/config/kafka-monitor.properties
@@ -39,7 +39,7 @@
     "zookeeper.connect": "localhost:2181",
     "bootstrap.servers": "localhost:9092",
     "produce.record.delay.ms": 100,
-    "produce.topic.autoTopicCreationEnabled": true,
+    "produce.topic.autoTopicCreationEnabled": false,
     "produce.topic.autoTopicReplicationFactor" : 1,
     "produce.producer.props": {
       "client.id": "kmf-client-id"
--- a/services/src/main/java/com/linkedin/kmf/common/Utils.java
+++ b/services/src/main/java/com/linkedin/kmf/common/Utils.java
@@ -9,27 +9,14 @@
  */
 package com.linkedin.kmf.common;
 
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
 import java.util.Properties;
-import java.util.Set;
-import kafka.cluster.Broker;
 import kafka.server.KafkaConfig;
 import org.apache.avro.generic.GenericData;
 import org.apache.avro.generic.GenericDatumWriter;
 import org.apache.avro.generic.GenericRecord;
 import org.apache.avro.io.Encoder;
 import org.apache.avro.io.JsonEncoder;
-import org.apache.kafka.clients.consumer.ConsumerConfig;
-import org.apache.kafka.clients.consumer.KafkaConsumer;
-import org.apache.kafka.common.Node;
-import org.apache.kafka.common.PartitionInfo;
 import org.apache.kafka.common.security.JaasUtils;
-import org.apache.kafka.common.serialization.StringDeserializer;
 import org.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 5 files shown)


services/src/main/java/com/linkedin/kmf/common/Utils.java
```diff
@@ -35,14 +40,54 @@
...
+   * @param topic topic name
+   * @param topicConfig if this is null or empty or min.insync.replicas is not specified then min.insync.replicas will
+   *                    be set to 3.
+   * @return the number of partitions created
+   */
+  public static int createMonitoringTopic(String zkUrl, String topic, Properties topicConfig) {
+    ZkUtils zkUtils = ZkUtils.apply(zkUrl, ZK_SESSION_TIMEOUT, ZK_CONNECTION_TIMEOUT, JaasUtils.isZkSecurityEnabled());
+    try {
+      //TODO: throw exception? check that we have the correct number of partitions?
+      if (AdminUtils.topicExists(zkUtils, topic)) {
```
lindong28: Is there any case where the topic will exist when we call this method AND we don't want to throw exception? If so, would it be better to name this method createTopicIfNotExist?

services/src/main/java/com/linkedin/kmf/common/Utils.java
```diff
@@ -35,14 +40,54 @@
...
+        LOG.warn("Monitoring topic \"" + topic + "\" already exists.");
+        return getPartitionNumForTopic(zkUrl, topic);
+      }
+
+      int brokerCount = zkUtils.getAllBrokersInCluster().size();
+      int partitionCount = brokerCount * 2;
+
+      // see min.insync.replicas
+      if (topicConfig == null) {
```
lindong28: It looks like topicConfig will always be an empty property as of current implementation. I am wondering if there any use-case where we need to override config for the topic. If not, maybe we can simplify the implementation by removing topicConfig this method.

services/src/main/java/com/linkedin/kmf/common/Utils.java
```diff
@@ -35,14 +40,54 @@
...
+        topicConfig = new Properties();
+      }
+      if (!topicConfig.contains("min.insync.replicas")) {
+        topicConfig.setProperty("min.insync.replicas", "" + 3);
```
lindong28: If replication factor is 3 and min.insync.replicas is 3, and if any broker goes down during rolling bounce, the availability will drop below 1 even though Kafka client is not negatively impacted. It is probably better to set min.insync.replicas to replication factor - 1. This allows us to capture event where 2 or more brokers are down simultaneously.

Why use "" + 3 instead simply "3"?

services/src/main/java/com/linkedin/kmf/common/Utils.java
```diff
@@ -35,14 +40,54 @@
...
+      }
+      AdminUtils.createTopic(zkUtils, topic, partitionCount, 3 /* replication factor*/, new Properties());
```
lindong28: We probably want to make replication factor configurable. In PROD environment we want to set replication factor to default replication factor used by most topics in this cluster. This value may vary across clusters. For example, it is currently 2 at LinkedIn. And we will increase it to 3 when we have JBOD setup. For open source users who want to Kafka Monitor in a cluster of 1 or 2 brokers, setting replication factor to 3 will cause topic creation to fail.

services/src/main/java/com/linkedin/kmf/common/Utils.java
smccauliff: I think it depends on the expected behavior.  If the default behavior that users expect is to manually create the topics then this warning makes sense or to have the topic already created then this warning makes sense.  In that case renaming the method as you propose makes it behavior more explicit. 

I will rename it.

Does it make sense to keep the current behavior and enable automatic topic creation via configuration?  Or would it be better to make automatic topic creation the new default behavior (this is what I was thinking).

services/src/main/java/com/linkedin/kmf/common/Utils.java
smccauliff: That sounds good.  I will make that change.

services/src/main/java/com/linkedin/kmf/common/Utils.java
smccauliff: See the next comment.

services/src/main/java/com/linkedin/kmf/common/Utils.java
smccauliff: Do we want to give the users the ability to configure the automatic topic creation with same level of detail provided by the command line?  The way of dealing with that is just to have the user configure the monitoring topic with the existing tools.   No?  Does having it in the configuration for the monitor better than having someone configure the topic externally?

-1 does sound like a better default for min.insync.replicas.  I will make that change.

Additional configuration properties are then:
* autoTopicCreationEnabled with default true
* autoTopicMinISR with default -1
* autoTopicReplicationFactor with default 3

And for future balancing:
* autoTopicPartitionMultiple with default 2.0
* autoTopicPartitionLowWaterMark with default 1.5

services/src/main/java/com/linkedin/kmf/common/Utils.java
```diff
@@ -27,6 +30,9 @@
 public class Utils {
   private static final Logger LOG = LoggerFactory.getLogger(Utils.class);
 
+  private static final int ZK_CONNECTION_TIMEOUT = 30_000;
```
lindong28: Should this be 3000 instead of 30_000? Would ZK_CONNECTION_TIMEOUT_MS and ZK_SESSION_TIMEOUT_MS be better names?

services/src/main/java/com/linkedin/kmf/services/configs/ProduceServiceConfig.java
```diff
@@ -63,6 +86,11 @@
                                     "kafka-monitor-topic",
                                     ConfigDef.Importance.MEDIUM,
                                     TOPIC_DOC)
+                            .define(CREATE_AUTO_TOPIC_CONFIG,
+                                    ConfigDef.Type.BOOLEAN,
+                                    false, //default
```
lindong28: Minor point: probably no need for comment here.

services/src/main/java/com/linkedin/kmf/services/configs/ProduceServiceConfig.java
```diff
@@ -49,6 +49,29 @@
...
+  public static final String REBALANCE_ENABLED_DOC = "Periodically move leaders and replica assignments so they are "
+      + "evenly spread amongst brokers.";
+
+  public static final String REBALANCE_PARTITION_MULTIPLE_CONFIG = "produce.topic.rebalancePartitionMultiple";
+  public static final String REBALANCE_PARTITION_MULTIPLE_DOC = "Determines the number of partitions per broker in the ideal case.";
+
+  public static final String REBALANCE_THRESHOLD_CONFIG = "produce.topic.rebalanceThreshold";
+  public static final String REBALANCE_THRESHOLD_DOC = "Determines the number of partitions per broker in the ideal case.";
+
+  public static final String CREATE_AUTO_TOPIC_CONFIG = "produce.topic.autoTopicCreationEnabled";
```
lindong28: Should it be named AUTO_TOPIC_CREATION_ENABLED_CONFIG?

config/kafka-monitor.properties
```diff
@@ -39,6 +39,9 @@
     "zookeeper.connect": "localhost:2181",
     "bootstrap.servers": "localhost:9092",
     "produce.record.delay.ms": 100,
+    "produce.topic.autoTopicCreationEnabled": true,
```
lindong28: With this config, when a new user of Kafka Monitor tries `./bin/kafka-monitor-start.sh config/kafka-monitor.properties` as suggested in our wiki on a cluster of only 1-2 machines, they will see topic creation failure without knowing that they need to change autoTopicReplicationFactor to make this work. This may be a frustrating experience. Maybe we should turn of autoTopicCreationEnabled here?

services/src/main/java/com/linkedin/kmf/common/Utils.java
lindong28: I think it will be better for new user who wants to test out Kafka Monitor to enable auto topic creation and set replication factor to 1 by default. Production user can override these configs to whatever value is appropriate for their cluster.

services/src/main/java/com/linkedin/kmf/common/Utils.java
lindong28: Thanks @smccauliff. I have commented on the latest patch. -1 is an invalid value for min.insync.replicas, right? Maybe we can simply remove this config and use replication factor - 1 as min.insync.replicas, set autoTopicCreationEnabled to true and set autoTopicReplicationFactor to 1 (just in case there is only one broker in the cluster).

config/kafka-monitor.properties
smccauliff: If the defaults are changed in kafka-monitor.properties then why do they need to be changed in BasicEndToEndTest.java?  Are these not the properties for the basic end to end test?

Yes, things should work as advertised.  I will make the other changes.

services/src/main/java/com/linkedin/kmf/common/Utils.java
smccauliff: Elsewhere it was the literal 30000. When I made it a constant I used the 30_000 format because I find it more readable than 30000.  I was not sure of the units for these parameters so I did not put that in the name of the constant.  I will add them to the name of the constant.

services/src/main/java/com/linkedin/kmf/services/configs/ProduceServiceConfig.java
smccauliff: Changed.

services/src/main/java/com/linkedin/kmf/services/configs/ProduceServiceConfig.java
```diff
@@ -92,7 +120,32 @@
                                     ConfigDef.Type.INT,
                                     5,
                                     ConfigDef.Importance.LOW,
-                                    PRODUCE_THREAD_NUM_DOC);
+                                    PRODUCE_THREAD_NUM_DOC)
+                            .define(AUTO_TOPIC_MIN_ISR_CONFIG,
+                                    ConfigDef.Type.INT,
+                                    -1,
```
smccauliff: OK.  I will remove the config and set min.isr to max(replicationFactor -1, 1) when creating the auto topic.

config/kafka-monitor.properties
lindong28: Yeah, they don't have to be changed in BasicEndToEndTest. I mentioned this only because in the previous patch they use different default value for autoTopicCreationEnabled.

services/src/main/java/com/linkedin/kmf/common/Utils.java
```diff
@@ -35,13 +54,178 @@
...
+      if (partitionFactor <= 0) {
+        throw new IllegalArgumentException("Partition factor must be greater than zero, but was configured for " +
+            partitionFactor + ".");
+      }
+      int partitionCount = brokerCount * partitionFactor;
+
+      int minIsr = Math.max(replicationFactor - 1, 1);
+      Properties topicConfig = new Properties();
+      topicConfig.setProperty(KafkaConfig.MinInSyncReplicasProp(), Integer.toString(minIsr));
+      AdminUtils.createTopic(zkUtils, topic, partitionCount, replicationFactor, new Properties());
```
lindong28: Should you use topicConfig as last parameter of this call?

services/src/main/java/com/linkedin/kmf/common/Utils.java
```diff
@@ -35,13 +54,178 @@
...
+   * This is so you can call KafkaConsumer.partitionsFor don't consume messages from this that is the job of the
+   * ConsumeService.
+   *
+   * @param brokerList probably get this from the configuration file.
+   * @param zkUrl zoo keeper url
+   * @param topic the monitoring topic
+   * @return an initialized KafkaConsumer
+   */
+
+  private static KafkaConsumer<?,?> constructClientForMetadata(String brokerList, String zkUrl, String topic) {
```
lindong28: This code is related to partition rebalance. Do you intend to get these reviewed in this pull request or are they committed here by accident?

config/kafka-monitor.properties
lindong28: Ah now I understand your question. Right, kafka-monitor.properties is not used by BasicEndToEndTest.java. It is only used by ./bin/kafka-monitor-start.sh  and KafkaMonitor.java as input.

services/src/main/java/com/linkedin/kmf/common/Utils.java
```diff
@@ -35,13 +54,178 @@
...
+      double partitionThreshold) {
+
+    if (partitionThreshold < 1) {
+      throw new IllegalArgumentException(
+          "partitionThreshold must be greater than or equal to one, but found " + partitionThreshold + ".");
+    }
+
+    ZkUtils zkUtils = ZkUtils.apply(zkUrl, ZK_SESSION_TIMEOUT_MS, ZK_CONNECTION_TIMEOUT_MS, JaasUtils.isZkSecurityEnabled());
+
+    try (KafkaConsumer<?, ?> consumerForMetadata = constructClientForMetadata(brokersList, zkUrl, topic)) {
```
lindong28: We probably don't need to construct a client to get partition info of a given topic. getPartitionsForTopics() from ZkUtils can read these from zookeeper. Can you try it?

smccauliff: I started down that route but in involves converting a scala data structure into java.  It seems somewhat complicated since it involves nested collections.  Specifically Map[Int, Seq[Int]] needs to be converted.  In the past I've had problems with using java.lang.Integer in collections when calling java methods that have arguments like Collection<Integer>.  If you know how to do this conversion and it's not going to take all day to figure it out then I'm open to using the equivalent functionality in ZkUtils.

services/src/main/java/com/linkedin/kmf/services/configs/ProduceServiceConfig.java
```diff
@@ -63,6 +83,10 @@
                                     "kafka-monitor-topic",
                                     ConfigDef.Importance.MEDIUM,
                                     TOPIC_DOC)
+                            .define(AUTO_TOPIC_CREATION_ENABLED_CONFIG,
+                                    ConfigDef.Type.BOOLEAN,
+                                    false,
```
smccauliff: I'll make it default in EndToEndTest only.

services/src/main/java/com/linkedin/kmf/common/Utils.java
smccauliff: Oops.  Yes.

services/src/main/java/com/linkedin/kmf/common/Utils.java
smccauliff: I'm going to remove this code from this pull request and we can discuss it in a different pull request.

services/src/main/java/com/linkedin/kmf/common/Utils.java
```diff
@@ -35,13 +54,178 @@
...
+    KafkaConsumer<String, String> metadataConsumer = new KafkaConsumer<>(consumerProps);
+    metadataConsumer.subscribe(Collections.singletonList(topic));
+    return metadataConsumer;
+  }
+
+  /**
+   *
+   * @param zkUrl the zookeeper url
+   * @param topic the monitored topic
+   * @param partitionThreshold  This assumes that you want to have at least one per broker.  When the number of partitions
```
smccauliff: I'm going to remove this code from this pull request and we can discuss it in a different pull request.

services/src/main/java/com/linkedin/kmf/common/Utils.java
```diff
@@ -35,13 +54,178 @@
...
+      return true;
+    }
+
+    Map<Integer, Integer> brokerToPartitionCount = new HashMap<>(brokers.size());
+    Set<Integer> leaders = new HashSet<>(brokers.size());
+
+    // Count the number of partitions a broker is involved with and if it is a leader for some partition
+    // Check that a partition has at least a certain number of replicas
+    for (PartitionInfo partitionInfo : partitionInfoList) {
+      if (partitionInfo.replicas().length < minPartitionCountPerBroker) {
```
smccauliff: I'm going to remove this code from this pull request and we can discuss it in a different pull request.

tests/src/main/java/com/linkedin/kmf/tests/BasicEndToEndTest.java
```diff
@@ -82,122 +83,122 @@ public void awaitShutdown() {
...
+    parser.addArgument("--producer-id").action(store()).required(false).type(String.class).dest("producerId").help("The producerId will be used by producer client and encoded in the messages to the topic");
 
     parser.addArgument("--broker-list")
-      .action(store())
-      .required(true)
-      .type(String.class)
-      .metavar("HOST1:PORT1[,HOST2:PORT2[...]]")
-      .dest("brokerList")
-      .help("Comma-separated list of Kafka brokers in the form HOST1:PORT1,HOST2:PORT2,...");
+        .action(store())
```
smccauliff: This is strange because I don't see this in my local copy of this file.

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
```diff
@@ -49,31 +56,57 @@
...
-  private final int _partitionNum;
+  /** This can be updated while running when new partitions are added to the monitored topic. */
+  private final ConcurrentMap<Integer, AtomicLong> _nextIndexPerPartition;
+  /** This is the last thing that should be updated after adding new partitions. */
+  private final AtomicInteger _partitionNum;
   private final int _recordSize;
   private final String _topic;
   private final String _producerId;
   private final AtomicBoolean _running;
+  private final String _zkConnect;
```
lindong28: Why do we need _zkConnect and _brokerList as private variable here? Is this needed later for rebalance related code?

tests/src/main/java/com/linkedin/kmf/tests/BasicEndToEndTest.java
```diff
@@ -82,122 +83,122 @@ public void awaitShutdown() {
...
-      .dest("topic")
-      .help("Produce messages to this topic and consume message from this topic");
-
-    parser.addArgument("--producer-id")
-      .action(store())
-      .required(false)
-      .type(String.class)
-      .dest("producerId")
-      .help("The producerId will be used by producer client and encoded in the messages to the topic");
+    ArgumentParser parser = ArgumentParsers.newArgumentParser("").defaultHelp(true).description("");
```
lindong28: Is there any reason to put them on the same line? Can we keep the same format as other `parser.addArgument(...)`?

tests/src/main/java/com/linkedin/kmf/tests/BasicEndToEndTest.java
lindong28: It is probably because the use of tab. Have you changed configured idea to use only tab for indentation?

It is probably better to keep the 2-space indentation if possible. This makes code style consistent and avoid large chunk of code change in the pull request.

services/src/main/java/com/linkedin/kmf/common/Utils.java
lindong28: Thanks!

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
```diff
@@ -49,31 +56,57 @@
...
     int threadsNum = config.getInt(ProduceServiceConfig.PRODUCE_THREAD_NUM_CONFIG);
     _topic = config.getString(ProduceServiceConfig.TOPIC_CONFIG);
     _producerId = config.getString(ProduceServiceConfig.PRODUCER_ID_CONFIG);
     _produceDelayMs = config.getInt(ProduceServiceConfig.PRODUCE_RECORD_DELAY_MS_CONFIG);
     _recordSize = config.getInt(ProduceServiceConfig.PRODUCE_RECORD_SIZE_BYTE_CONFIG);
     _sync = config.getBoolean(ProduceServiceConfig.PRODUCE_SYNC_CONFIG);
-    _partitionNum = Utils.getPartitionNumForTopic(zkConnect, _topic);
+    _partitionNum = new AtomicInteger(0);
+
+    if (_rebalanceThreshold < 1) {
```
lindong28: We can add the `atLeast` in the next pull request if you like. I am just thinking we don't need the code here to check if rebalanceThreshold is < 1.

smccauliff: Cool.  I wasn't aware of that.  Yes, can we please fix that in a different pull request.

tests/src/main/java/com/linkedin/kmf/tests/BasicEndToEndTest.java
smccauliff: I sent you an email about this.

tests/src/main/java/com/linkedin/kmf/tests/BasicEndToEndTest.java
smccauliff: I must have accidentally hit something in IntelliJ

PR #23, opened 2016-10-05
add a Docker build folder with makefile

coffeepac:
make this an image my friends!

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -14,8 +14,8 @@
 
 all: push
 
-TAG = 0.1
-PREFIX = quay.io/coffeepac/kafka-monitor
+TAG ?= 0.1
+PREFIX ?= quay.io/samsung_cnct/kafka-monitor
 
 container:
 	cd ../ && \
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -1,0 +1,28 @@
+# Copyright 2016 The Kubernetes Authors All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


lindong28:
Thanks @coffeepac. Can you add Apache License 2.0 to all files?

coffeepac:
@lindong28 added license files, a small CMD update and a README.  Anything else?

Thanks for the quick turnaround as well!

docker/Makefile
```diff
@@ -0,0 +1,28 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+all: push
+
+TAG = 0.1
+PREFIX = quay.io/coffeepac/kafka-monitor
```
lindong28: I should have noted this previously. Just a minor comment: can we change the default location to something general? coffeepac is your username and doesn't seem very general. Is there any specific meaning of quay.io, maybe a common name among docker user?

coffeepac: Fair!  I meant to ask about this.  quay.io is a public image repo, operated by coreos.  It's what my team uses. So nothing special about it.  

Does linkedin have a normal location?  Im happy to write it there.  linkedin/kafka-monitor or something?

coffeepac: quay.io uses GitHub for auth and namespaces based on GitHub user ID.  I don't think we can push to /default/ like that.  If I can't do something like you suggest do you have a group GitHub account you can make a quay repo for?  Or I can push it to somewhere under quay.io/samsung_cnct/

I'll take a look at having the makefile do the bashy have a default value if not set thing.  Should be just one line.

coffeepac:
@lindong28 updated the default repo and the README with instructions in how to use the makefile to control the docker build with command line flags or environment variables

lindong28:
Thanks @coffeepac!

PR #24, opened 2016-10-06
Use testng on the tests project and remove dependencies on junit. 

smccauliff:
Enable the test configuration on the tests project.
    Remove dependency on junit as it has a conflicting license and kafka monitor does not use it anyway.

 As there are not any unit tests anyway this should not have an negative impact and will allow others to write unit tests.


lindong28:
Thanks for the patch @smccauliff

PR #25, opened 2016-10-13
Monitored topic rebalancing.

smccauliff:
This implements: adding partitions, reassigning partitions, and preferred leader election when the monitored topic falls out of its preferred configuration.


services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
```diff
@@ -0,0 +1,427 @@
...
+import static com.linkedin.kmf.common.Utils.ZK_SESSION_TIMEOUT_MS;
+
+import scala.collection.Seq;
+
+
+/**
+ * Runs this periodically to rebalance the monitored topic across brokers and to reassign leaders to brokers so that the
+ * monitored topic is sampling all the brokers evenly.
+ */
+class TopicRebalancer implements Runnable {
```
lindong28: It seems that the TopicRebalancer is itself a periodic task that can run independently of ProduceServce. How about we decouple it from ProduceService and have a standalone TopicBalanceService so that user can specify it in kafka-monitor.properties when they start KafkaMonitor? This allows code to be simpler (by decoupling TopicRebalanceService from ProduceService) and config to be simpler (by removing produce.topic.rebalanceEnabled and put all rebalance related config under RebalanceServiceConfig)

services/src/main/java/com/linkedin/kmf/common/Utils.java
```diff
@@ -34,8 +34,8 @@
 public class Utils {
   private static final Logger LOG = LoggerFactory.getLogger(Utils.class);
 
-  private static final int ZK_CONNECTION_TIMEOUT_MS = 30_000;
-  private static final int ZK_SESSION_TIMEOUT_MS = 30_000;
+  public static final int ZK_CONNECTION_TIMEOUT_MS = 30_000;
+  public static final int ZK_SESSION_TIMEOUT_MS = 30_000;
```
lindong28: Can we also simplify the code in method `createMonitoringTopicIfNotExists()` by removing check for method `partitionFactor <= 0`? We can validate the config in XXXConfig class. And `AdminUtils.createTopic(...)` will throw exception if partition number <= 0.

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
lindong28: One concern is that we currently need to restart `ProduceService` if we increase number of partition for the topic. Otherwise ProduceService will still only produce to old partitions. I talked with SRE and they told me they typically expend Kafka cluster once every quarter. The frequency of doubling the size of Kafka cluster will be even lower. They are OK with manually restarting Kafka Monitor every time they increase the size of Kafka cluster. Therefore we can make the code simpler by not having to restart ProduceService.

services/src/main/java/com/linkedin/kmf/services/configs/ProduceServiceConfig.java
```diff
@@ -57,16 +59,19 @@
   public static final String REBALANCE_ENABLED_DOC = "Periodically move leaders and replica assignments so they are "
       + "evenly spread amongst brokers.";
 
-  public static final String REBALANCE_PARTITION_MULTIPLE_CONFIG = "produce.topic.rebalancePartitionMultiple";
-  public static final String REBALANCE_PARTITION_MULTIPLE_DOC = "Determines the number of partitions per broker in the ideal case.";
+  public static final String REBALANCE_PARTITION_FACTOR_CONFIG = "produce.topic.rebalancePartitionFactor";
+  public static final String REBALANCE_PARTITION_FACTOR_DOC = "Determines the number of partitions per broker in the ideal case.";
```
lindong28: I find it a bit hard to interpret "ideal" here. This is actually the default number of partitions when we create/expand the partitions of a topic, right? Maybe we can say something like "The default ratio of partitions to broker when RebalanceService create/expand partitions of a topic".

I am also thinking that we can allow the config name to explain itself better -- how about "rebalance.default.partitionToBrokerRatio"

services/src/main/java/com/linkedin/kmf/services/configs/ProduceServiceConfig.java
```diff
@@ -57,16 +59,19 @@
...
 
   public static final String REBALANCE_THRESHOLD_CONFIG = "produce.topic.rebalanceThreshold";
   public static final String REBALANCE_THRESHOLD_DOC = "Determines the number of partitions per broker in the ideal case.";
 
+  public static final String REBALANCE_DELAY_MS_CONFIG = "produce.topic.rebalanceDelayMs";
```
lindong28: Is there any reason that user wants to explicitly config this value? If not, maybe we can simply provide a default value without exposing this as a config.

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
```diff
@@ -0,0 +1,427 @@
...
+   *    increment _PartitionNum (callback does this)
+   * 2. This is done during rebalance.
+   *   if a broker does not have enough partitions then assign it partitions from brokers that have excess
+   *   if a broker is not a leader of some partition then make it a leader of some partition by finding out
+   *     which broker is a leader of more than one partition.  Does this involve moving partitions among brokers?
+   * 3. run a preferred replica election
+   * </pre>
+   */
+  void rebalanceMonitoredTopic(RebalanceCondition rebalanceCondition) {
+    ZkUtils zkUtils = createZooKeeperUtils();
```
lindong28: It seems that we zkUtils will be created exactly once. Can we simply create it once in the constructor? If we are concerned that this is a blocking call, then we can create it in the method `Service.start()`.

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
```diff
@@ -0,0 +1,427 @@
...
+   * @param partitionInfoList this should only contain partition info for the monitored topic
+   * @param brokers get this from ZkUtils, this should be all the active brokers in the cluster
+   * @param partitionThreshold the lower water mark for when we do not have enough monitored partitions
+   * @return see above
+   */
+  static RebalanceCondition monitoredTopicNeedsRebalance(List<PartitionInfo> partitionInfoList, Collection<Broker> brokers,
+      double partitionThreshold) {
+
+    int partitionCount = partitionInfoList.size();
+    int minPartitionCount = (int) (brokers.size() * partitionThreshold);
```
lindong28: Can you explain why we need to round the number of an integer here? I think user expects a rebalance if partition_num/broker_num < minimum_partition_to_broker_ratio, right?

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
```diff
@@ -0,0 +1,427 @@
...
+    return partitionInfoList;
+  }
+
+
+  /**
+   * This returns RebalanceCondition.OK unless one of the following conditions are met in this order:
+   * <ul>
+   *   <li> The minimum number of total monitored partitions falls below floor(brokers * partitionThreshold).
+   *        Returns PartitionsLow</li>
+   *   <li> One or more brokers does not have a monitored partition or falls below the minimum number of monitored
```
lindong28: I am not sure we need to check that every broker has at least a minimum number of partitions. We can discuss offline.

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
```diff
@@ -0,0 +1,427 @@
...
+ * monitored topic is sampling all the brokers evenly.
+ */
+class TopicRebalancer implements Runnable {
+
+  private static final Logger LOG = LoggerFactory.getLogger(TopicRebalancer.class);
+
+  /**
+   * The state of the monitored topic.
+   */
+  enum RebalanceCondition {
```
lindong28: This kind of imply that these 4 conditions are mutually exclusive, but they are not. For example, we can have a case where a broker doesn't have any partition and thus satisfied the last 3 conditions.

And alternative is to have the following logic in run():
```
if (isPartitionLow) increasePartitionNum()
if (isAnyBrokerNotPreferredLeader()) reassignReplicas()
if (isAnyBrokerNotElectedLeader()) doPreferredLeaderElection()
```

config/kafka-monitor.properties
```diff
@@ -51,6 +52,16 @@
     }
   },
 
+  "topic-rebalancer": {
```
lindong28: To make the name consistent with other services, how about naming it "topic-rebalance-service"?

tests/src/main/java/com/linkedin/kmf/tests/BasicEndToEndTest.java
```diff
@@ -83,133 +84,151 @@ public void awaitShutdown() {
...
+
+    parser.addArgument("--topic-rebalance-enabled")
+      .action(store())
+      .required(false)
+      .type(Boolean.class)
+      .metavar("REBALANCE_ENABLED")
+      .dest("rebalanceEnabled")
+      .help("When true this creates and starts a topic rebalancer.");
+
+    parser.addArgument("--topic-rebalance-ms")
```
lindong28: I am not sure "topic-rebalance-ms" tells user what is means. How about "topic-rebalance-interval-ms"?

tests/src/main/java/com/linkedin/kmf/tests/BasicEndToEndTest.java
```diff
@@ -83,133 +84,151 @@ public void awaitShutdown() {
...
+      .help("The granularity in ms of latency percentile metric. This is the width of the bucket used in percentile calculation.");
 
     parser.addArgument("--auto-topic-creation")
-        .action(store())
-        .required(false)
-        .type(Boolean.class)
-        .metavar("AUTO_TOPIC_CREATION_ENABLED")
-        .dest("autoTopicCreationEnabled")
-        .help(ProduceServiceConfig.AUTO_TOPIC_CREATION_ENABLED_DOC);
+      .action(store())
```
lindong28: How about name it "topic-creation-enabled"?

services/src/main/java/com/linkedin/kmf/services/configs/TopicRebalancerConfig.java
```diff
@@ -0,0 +1,77 @@
...
+  public static final String REBALANCE_EXPECTED_RATIO_CONFIG = "rebalance.expectedPartitionBrokerRatio";
+  public static final String REBALANCE_EXPECTED_RATIO_DOC = "The expected ratio of partition / broker.  When the actual ratio falls below this new partitions are created.";
+
+  public static final String REBALANCE_INTERVAL_MS_CONFIG = "rebalance.intervalMs";
+  public static final String REBALANCE_INTERVAL_MS_DOC = "The gap in ms between the times the cluster balance on the monitored topic is checked.";
+
+  public static final String PARTITIONS_PER_BROKER_CONFIG = ProduceServiceConfig.PARTITIONS_PER_BROKER_CONFIG;
+  public static final String PARTITIONS_PER_BROKER_DOC = ProduceServiceConfig.PARTITIONS_PER_BROKER_DOC;
+
+  public static final String TOPIC_REPLICATION_FACTOR_CONFIG = ProduceServiceConfig.TOPIC_REPLICATION_FACTOR_CONFIG;
```
lindong28: I think the job of rebalance service is to increase partition number or reassign partitions. The decision of replication factor should stay with whoever create the topic. Therefore this service probably use the existing replication factor.

services/src/main/java/com/linkedin/kmf/services/configs/TopicRebalancerConfig.java
```diff
@@ -0,0 +1,77 @@
...
+  public static final String ZOOKEEPER_CONNECT_CONFIG = CommonServiceConfig.ZOOKEEPER_CONNECT_CONFIG;
+  public static final String ZOOKEEPER_CONNECT_DOC = CommonServiceConfig.ZOOKEEPER_CONNECT_DOC;
+
+  public static final String TOPIC_CONFIG = CommonServiceConfig.TOPIC_CONFIG;
+  public static final String TOPIC_DOC = CommonServiceConfig.TOPIC_DOC;
+
+  public static final String REBALANCE_EXPECTED_RATIO_CONFIG = "rebalance.expectedPartitionBrokerRatio";
+  public static final String REBALANCE_EXPECTED_RATIO_DOC = "The expected ratio of partition / broker.  When the actual ratio falls below this new partitions are created.";
+
+  public static final String REBALANCE_INTERVAL_MS_CONFIG = "rebalance.intervalMs";
```
lindong28: To stay inline with the config naming conversion in Kafka, this should probably be named rebalance.interval.ms.

services/src/main/java/com/linkedin/kmf/services/configs/ProduceServiceConfig.java
```diff
@@ -49,24 +52,17 @@
...
-  public static final String REBALANCE_ENABLED_CONFIG = "produce.topic.rebalanceEnabled";
-  public static final String REBALANCE_ENABLED_DOC = "Periodically move leaders and replica assignments so they are "
-      + "evenly spread amongst brokers.";
-
-  public static final String REBALANCE_PARTITION_MULTIPLE_CONFIG = "produce.topic.rebalancePartitionMultiple";
-  public static final String REBALANCE_PARTITION_MULTIPLE_DOC = "Determines the number of partitions per broker in the ideal case.";
-
-  public static final String REBALANCE_THRESHOLD_CONFIG = "produce.topic.rebalanceThreshold";
-  public static final String REBALANCE_THRESHOLD_DOC = "Determines the number of partitions per broker in the ideal case.";
+  public static final String PARTITIONS_PER_BROKER_CONFIG = "topic.partitionsPerBroker";
```
lindong28: By convention we prefix the config name and the variable name with the service name, e.g. "produce.producer.props". The goal is to avoid config name collision between different services and make it easy for user to know which service is the config applied to. Therefore we probably want to change the name for "topic.replicationFactor" and "topic.partitionsPerBroker" as appropriate.

Previously we discussed which service should take care of topic creation, ProduceService or TopicRebalanceService. Now I look at the patch, it seems cleaner to do it in the TopicRebalanceService. Otherwise we need to define partitionToBrokerRatio in multiple services, which seems ugly. It seems reasonable to have one module that manages both topic rebalance and topic creation.

It is true that produce service won't be able to run if topic is not created and rebalance service is not used. But if rebalance service is explicitly disabled (suppose we enable it by default), it means user wants to manually manage the number of partitions in case of cluster expansion, which in turn indicates that they are OK with manually creating topic.

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
```diff
@@ -223,7 +236,11 @@ public double measure(MetricConfig config, long now) {
             double availabilitySum = 0.0;
             int partitionNum = _partitionNum.get();
             for (int partition = 0; partition < partitionNum; partition++) {
-              double recordsProduced = _sensors.metrics.metrics().get(new MetricName("records-produced-rate-partition-" + partition, METRIC_GROUP_NAME, tags)).value();
+              MetricName partitionRecordsProduced = new MetricName("records-produced-rate-partition-" + partition, METRIC_GROUP_NAME, tags);
+              if (!_sensors.metrics.metrics().containsKey(partitionRecordsProduced)) {
```
lindong28: I am not sure that we need this check. `addPartitionSensors(...)` should have created sensors for all partitions, right?

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
```diff
@@ -284,4 +301,41 @@ public void run() {
     }
   }
 
+
+  private class CheckForNewPartitons implements Runnable {
+
+    public void run() {
```
lindong28: Some code seems redundant here. The logic for initializing should be in `start()` and the logic for clearing up state (e.g. close executor) should be in `close()`. The method should probably look something like this:

```
if (Utils.getPartitionNumForTopic(zkUrl, topic) > _partitionNum) {
  close();
  start();
}
```

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
```diff
@@ -0,0 +1,427 @@
...
+import static com.linkedin.kmf.common.Utils.ZK_SESSION_TIMEOUT_MS;
+
+import scala.collection.Seq;
+
+
+/**
+ * Runs this periodically to rebalance the monitored topic across brokers and to reassign _electedLeaders to brokers so that the
+ * monitored topic is sampling all the brokers evenly.
+ */
+public class TopicRebalancer implements Service  {
```
lindong28: Can we name it TopicRebalanceService? We usually have Service at the end. This is similar to `ProduceService` and `ConsumerService`.

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
```diff
@@ -0,0 +1,427 @@
...
+      this._allBrokers = allBrokers;
+      this._partitionInfo = partitionInfo;
+    }
+
+    /**
+     *
+     * @param expectedRatio the expected partition to broker ratio
+     * @return true if the current partition to broker ratio is less than the expected ratio
+     */
+    boolean partitionsLow(double expectedRatio) {
```
lindong28: I think it is better for `partitionsLow`, `brokerMissingPartition` and `brokerNotElectedLeader` to have the same style. Eitehr they are all stateful (i.e. read state from class private variables ) or stateless (i.e. read state from method input). Maybe we can change the method signature to be `partitionLow()`

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
```diff
@@ -0,0 +1,427 @@
...
+    }
+
+  }
+
+  private class TopicRebalancerRunnable implements Runnable {
+    @Override
+    public void run() {
+      try {
+        TopicState topicState = topicState();
+        if (topicState.brokerNotElectedLeader() || topicState.brokerMissingPartition() ||
```
lindong28: It seems unnecessary to check violations before calling `rebalanceMonitoredTopic(...)`. We can simply do the following in this method, which roughly move the body of `rebalanceMonitoredTopic(...)` here:


```
try {
  if (topicState.partitionsLow())
    doSomething();
  if (topicState.brokerMissingPartition())
     doSomething();
  if (topicState.brokerNotElectedLeader())
     doSomething();
} catch (Exception e) {
  // Assume this is not a retraible exception.
  LOG.error(....);
  close();
}
```

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
```diff
@@ -0,0 +1,427 @@
...
+    _zkUtils = ZkUtils.apply(zkUrl, ZK_SESSION_TIMEOUT_MS + _scheduleIntervalMs, ZK_CONNECTION_TIMEOUT_MS + _scheduleIntervalMs,
+      JaasUtils.isZkSecurityEnabled());
+    _executor = Executors.newScheduledThreadPool(1);
+
+    LOG.info("Topic rebalancer service \"" + _serviceName + "\" constructed with expected partition/broker ratio " + _expectedPartitionBrokerRatio +
+      " topic " + _topic + " _partitionsPerBroker " + _partitionsPerBroker + " scheduleIntervalMs " +
+      _scheduleIntervalMs + " replication factor " + _replicationFactor + ".");
+  }
+
+  int scheduleDurationMs() {
```
lindong28: This method is not used. Can we remove it?

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
```diff
@@ -0,0 +1,427 @@
...
+  }
+
+  protected void addPartitions(ZkUtils zkUtils, int addPartitionCount) {
+    AdminUtils.addPartitions(zkUtils, _topic, addPartitionCount, null, false);
+  }
+
+  /**
+   * @return the state of the topic e.g. all the partitions and replicas
+   */
+  private TopicState topicState() {
```
lindong28: Can we name it `getTopicState()`?

services/src/main/java/com/linkedin/kmf/services/configs/TopicRebalancerConfig.java
```diff
@@ -0,0 +1,77 @@
...
+
+  private static final ConfigDef CONFIG;
+
+  public static final String ZOOKEEPER_CONNECT_CONFIG = CommonServiceConfig.ZOOKEEPER_CONNECT_CONFIG;
+  public static final String ZOOKEEPER_CONNECT_DOC = CommonServiceConfig.ZOOKEEPER_CONNECT_DOC;
+
+  public static final String TOPIC_CONFIG = CommonServiceConfig.TOPIC_CONFIG;
+  public static final String TOPIC_DOC = CommonServiceConfig.TOPIC_DOC;
+
+  public static final String REBALANCE_EXPECTED_RATIO_CONFIG = "rebalance.expectedPartitionBrokerRatio";
```
smccauliff: Expected means that this is what one would expect as opposed to what one actually has.

services/src/main/java/com/linkedin/kmf/services/configs/TopicRebalancerConfig.java
smccauliff: OK.

services/src/main/java/com/linkedin/kmf/services/configs/TopicRebalancerConfig.java
smccauliff: How does one find out the replication factor of the topic?  My understanding is that a partition need not have the same replication factor as the other partitions in the same topic.

tests/src/main/java/com/linkedin/kmf/tests/BasicEndToEndTest.java
smccauliff: OK.

tests/src/main/java/com/linkedin/kmf/tests/BasicEndToEndTest.java
smccauliff: OK.

config/kafka-monitor.properties
smccauliff: OK.

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
```diff
@@ -41,61 +44,60 @@
...
- */
 public class ProduceService implements Service {
   private static final Logger LOG = LoggerFactory.getLogger(ProduceService.class);
   private static final String METRIC_GROUP_NAME = "produce-service";
 
   private final String _name;
   private final ProduceMetrics _sensors;
-  private final KMBaseProducer _producer;
-  private final ScheduledExecutorService _executor;
+  private volatile KMBaseProducer _producer;
```
smccauliff: The producer and the executor service is recreated when new partitions are detected.  I often make members volatile when I expect them to be updated and read by different threads.  Since new threads are created when the producer is changed out probably this is not needed.  But it is also unlikely to affect performance.

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
```diff
@@ -0,0 +1,427 @@
...
+      if (_zkUtils.getPartitionAssignmentForTopics(scalaTopic).apply(_topic).size()  == expectedNumberOfPartitions) {
+        return;
+      }
+      try {
+        Thread.sleep(1000);
+      } catch (InterruptedException e) {
+        throw new IllegalStateException(e);
+      }
+    }
+    throw new IllegalStateException("Waiting for additional partitions to appear timed out.");
```
lindong28: It maybe more informative to say "Partition expansion can not finish in 60 sec". Besides, I am not sure if it is appropriate to use IllegalStateException here.

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
```diff
@@ -0,0 +1,427 @@
...
+public class TopicRebalancer implements Service  {
+
+  private static final Logger LOG = LoggerFactory.getLogger(TopicRebalancer.class);
+
+  /**
+   * The state of the monitored topic.
+   */
+  static class TopicState {
+
+    Map<Integer, Integer> _brokerToPartitionCount = new HashMap<>();
```
lindong28: It seems that we don't need `_brokerToPartitionCount`. The only time we need it is in brokerMissingPartition(), where we check `_brokerToPartitionCount.containsKey(broker.id())`. But this check is already covered by `_preferredLeaders.contains(broker.id())`

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
```diff
@@ -41,61 +44,60 @@
...
+  private volatile ScheduledExecutorService _executor;
+  private final ScheduledExecutorService _checkAddPartitionsExecutor;
```
smccauliff: I tried using the existing thread pool and having the thread coordinate with each other.  There is a way to make that correct but it is complicated by:
1) only wanting one of N threads to check periodically
2) coordinating with all the running threads
3) checking if we need a new CyclicBarrier

1) We basically need to recreate what the schedule thread executor provides already, but we want one of the scheduled threads to do this at a potentially different interval;  you probably don't want to poll zookeeper every 100ms. This needs to set some flag for threads to wait on the barrier (see 2).
2) All the threads need to stop so that the producer can be swapped out with a new producer.  This means introducing a CyclicBarrier.  When all the expected threads hit the barrier the Runnable associated with the barrier will trigger which will change the producer.
3) The CyclicBarrier does not actually trigger on a fixed number of threads.  The number of active threads is not actually equal to the number of threads in the executor service.  It's actually min(number of partitions, number of threads).  So the cyclic barrier needs to be switched out.

Just using a separate thread in a different executor service that is scheduled to check for added partitions saves all this complexity.

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
```diff
@@ -0,0 +1,427 @@
...
+    //Assigned _electedLeaders tracks if there is some partition for which it is the preferred leader which could be none.
+    Set<Integer> _preferredLeaders = new HashSet<>();
+    Set<Integer> _electedLeaders = new HashSet<>();
+    Collection<Broker> _allBrokers;
+    List<PartitionInfo> _partitionInfo;
```
lindong28: This `_partitionInfo` can be replaced by `_partitionNum`.

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
smccauliff: It was handy for debugging.  I can remove it.

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
smccauliff: OK

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
smccauliff: I like to separate out the detection of a problem from the action that is taken.

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
smccauliff: OK

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
smccauliff: Why use "get"?  It does not mean anything unless you need to comply with Java Beans.  Scala does not use it.  Java itself has moved away from it.   For example in java collections it's "size()" not "getSize()".

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
```diff
@@ -0,0 +1,427 @@
...
+ * monitored topic is sampling all the brokers evenly.
+ */
+public class TopicRebalancer implements Service  {
+
+  private static final Logger LOG = LoggerFactory.getLogger(TopicRebalancer.class);
+
+  /**
+   * The state of the monitored topic.
+   */
+  static class TopicState {
```
lindong28: We actually only need one instance of TopicState, right? If so, would it be simpler to remove this class and only keep the states (e.g. _electedLeaders, broker list) as local variables of this service class? We only need a method like `readBrokerAndTopicInfo` when we need to get the latest state.

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
```diff
@@ -0,0 +1,427 @@
...
+     */
+    boolean partitionsLow(double expectedRatio) {
+      double actualRatio = ((double) _partitionInfo.size()) / _allBrokers.size();
+      return actualRatio < expectedRatio;
+    }
+
+    /**
+     * @return true if broker is not the preferred leader for at least one partition.
+     */
+    boolean brokerMissingPartition() {
```
lindong28: Maybe name it `anyBrokerNotPreferredLeader`.

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
```diff
@@ -0,0 +1,427 @@
...
+          return true;
+        }
+      }
+      return false;
+    }
+
+    /**
+     * @return true if at least one broker is not elected leader for at least one partition.
+     */
+    boolean brokerNotElectedLeader() {
```
lindong28: How about naming it `anyBrokerNotElectedLeader`?

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
smccauliff: Actually let's talk about what you want for this.

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
```diff
@@ -0,0 +1,427 @@
...
+      }
+    } catch (InterruptedException ie) {
+      throw new IllegalStateException(ie);
+    }
+  }
+
+  /**
+   *  Polling...
+   */
+  private void waitForAddedPartitionsToBecomeActive(int expectedNumberOfPartitions) {
```
lindong28: Since we have a method called `waitForOtherAssignmentsToComplete`, how about naming this method `waitForPartitionAdditionToComplete`?

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
smccauliff: OK

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
```diff
@@ -0,0 +1,427 @@
...
+      try {
+        LOG.debug("Waiting for current partition assignment to be complete.");
+        Thread.sleep(1000);
+      } catch (InterruptedException e) {
+        throw new IllegalStateException(e);
+      }
+    }
+  }
+
+  private void waitForPartitionReassignmentToComplete() throws InterruptedException {
```
lindong28: It is probably good enough to simply check if there is ongoing partition reassignment without looking at the topic name. The reason is that we only begin topic reassignment if there is no ongoing topic reassignment. In the rare case there someone else initiated a topic reassignment after our check but before we create our reassignment, it should be OK to wait for that topic as well.

Thus we can probably replace this method with waitForOtherAssignmentsToComplete(). Maybe rename it `waitForOngoingAssignmentsToComplete`

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
```diff
@@ -0,0 +1,427 @@
...
+    }
+    bldr.setLength(bldr.length() - 2);
+    bldr.append("]}");
+    return bldr.toString();
+  }
+
+  /**
+   * This runs the preferred replica election for all partitions.
+   */
+  protected void runPreferredElection(ZkUtils zkUtils, List<PartitionInfo> partitionInfoList) {
```
lindong28: It may be better named `triggerPreferredLeaderElection`?

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
smccauliff: I'll remove the volatile.

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
smccauliff: Can use number of partitions instead of partition info.

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
smccauliff: OK

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
smccauliff: Just leave it.

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
smccauliff: OK

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
smccauliff: What exception should be used?  It's a good runtime exception to use.

services/src/main/java/com/linkedin/kmf/services/TopicRebalancer.java
smccauliff: In order to detect the replication factor the partitionInfo is needed.

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
```diff
@@ -43,71 +43,89 @@
...
   private static final Logger LOG = LoggerFactory.getLogger(ProduceService.class);
   private static final String METRIC_GROUP_NAME = "produce-service";
 
   private final String _name;
   private final ProduceMetrics _sensors;
-  private final KMBaseProducer _producer;
-  private final ScheduledExecutorService _executor;
+  private KMBaseProducer _producer;
+  private ScheduledExecutorService _executor;
+  private final ScheduledExecutorService _checkAddPartitionsExecutor;
```
lindong28: Sorry if I am paranoid about the name here -- would it be better to name it handleNewPartitionsExecutor, given that the runnable is named HandleNewPartitions?

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
```diff
@@ -43,71 +43,89 @@
...
   private final AtomicInteger _partitionNum;
   private final int _recordSize;
   private final String _topic;
   private final String _producerId;
   private final AtomicBoolean _running;
-  private final String _zkConnect;
   private final String _brokerList;
-  private final double _rebalanceThreshold;
+  private final Map _producerPropsOverride;
+  private final String _producerClass;
```
lindong28: It seems unnecessary to have this variable _producerClass. We always use it together with _producerPropsOverride, which already contains _producerClass. Can we remove this variable and simplified the signature of method initializeProducer?

And it seems that the `_brokerList` and `_zkConnect` which were added in pull request https://github.com/linkedin/kafka-monitor/pull/22 doesn't need to be private variable as well, right?

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
```diff
@@ -43,71 +43,89 @@
...
     _sync = config.getBoolean(ProduceServiceConfig.PRODUCE_SYNC_CONFIG);
     _partitionNum = new AtomicInteger(0);
+    _running = new AtomicBoolean(false);
+    _nextIndexPerPartition = new ConcurrentHashMap<>();
 
-    if (_rebalanceThreshold < 1) {
-      throw new IllegalArgumentException("Rebalance threshold must be greater than one but is set to " + _rebalanceThreshold + ".");
-    }
-
+    int autoTopicPartitionFactor = config.getInt(ProduceServiceConfig.PARTITIONS_PER_BROKER_CONFIG);
```
lindong28: Can we name these two variables according to their config name?

services/src/main/java/com/linkedin/kmf/services/configs/TopicManagementConfig.java
```diff
@@ -0,0 +1,69 @@
...
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.linkedin.kmf.services.configs;
+
+import java.util.Map;
+import org.apache.kafka.common.config.AbstractConfig;
+import org.apache.kafka.common.config.ConfigDef;
+import static org.apache.kafka.common.config.ConfigDef.Range.atLeast;
+
+public class TopicManagementConfig extends AbstractConfig {
```
lindong28: Can we name it TopicManagementServiceConfig?

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
```diff
@@ -43,71 +43,89 @@
...
+
+    MetricConfig metricConfig = new MetricConfig().samples(60).timeWindow(1000, TimeUnit.MILLISECONDS);
+    List<MetricsReporter> reporters = new ArrayList<>();
+    reporters.add(new JmxReporter(JMX_PREFIX));
+    Metrics metrics = new Metrics(metricConfig, reporters, new SystemTime());
+    Map<String, String> tags = new HashMap<>();
+    tags.put("name", _name);
+    _sensors = new ProduceMetrics(metrics, tags);
+
+    _topicRebalanceService = new TopicManagementService(props, _name + "-topic-management-service");
```
lindong28: Now that we create the topic inside ProduceService we don't have to start TopicManagementService inside ProduceService, right? Besides, the variable name should probably be updated to something like `_topicManagementService`.

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
```diff
@@ -132,43 +150,37 @@ public ProduceService(Map<String, Object> props, String name) throws Exception {
...
 
   @Override
   public void start() {
     if (_running.compareAndSet(false, true)) {
-      int partitionNum = _partitionNum.get();
-      for (int partition = 0; partition < partitionNum; partition++) {
-        scheduleProduceRunnable(partition);
-      }
+      scheduleProduceRunnables();
+      _checkAddPartitionsExecutor.scheduleWithFixedDelay(new HandleNewPartitions(), 40_000, 40_000, TimeUnit.MILLISECONDS);
```
lindong28: Should we use the config topic-management.rebalance.interval.ms?

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
```diff
@@ -284,4 +300,47 @@ public void run() {
...
+        _executor.awaitTermination(Integer.MAX_VALUE, TimeUnit.MILLISECONDS);
+      } catch (InterruptedException e) {
+        throw new IllegalStateException(e);
+      }
+      _producer.close();
+      int nextPartitionStart = _partitionNum.get();
+      int addedPartitionCount = currentPartitionCount - nextPartitionStart;
+      int newPartitionCount = nextPartitionStart + addedPartitionCount;
+      for (int newPartition = nextPartitionStart; newPartition < newPartitionCount; newPartition++) {
+        LOG.debug(_name + ": Adding " + newPartition + " to sensors.");
```
lindong28: Since we change the partition number and add sensors for new partitions here, we should probably move following logic for add default index for new partitions from scheduleProduceRunnables() to here. What do you think?

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -0,0 +1,419 @@
...
+   * The state of the monitored topic.
+   */
+  static class TopicState {
+
+    final List<PartitionInfo> _partitionInfo;
+    //Assigned _electedLeaders tracks if there is some partition for which it is the preferred leader which could be none.
+    final Set<Integer> _preferredLeaders;
+    final Set<Integer> _electedLeaders;
+    final Collection<Broker> _allBrokers;
+    final double _expectedRatio;
```
lindong28: Maybe use the same name as the config such as partitionToBrokerRatioThreshold?

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -0,0 +1,419 @@
...
+    LOG.info("Topic management service \"" + _serviceName + "\" constructed with expected partition/broker ratio " + _partitionBrokerRatioThreshold
+      + " topic " + _topic + " partitionsPerBroker " + _partitionsPerBroker + " scheduleIntervalMs " +
+      _scheduleIntervalMs + ".");
+  }
+
+
+  /**
+   *  Polling...
+   */
+  private void waitForAddedPartitionsToBecomeActive(int expectedNumberOfPartitions) {
```
lindong28: It seems that this method is not needed because `AdminUtils.addPartitions(..)` will update the same znode that this method reads.

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -0,0 +1,419 @@
...
+    for (PartitionInfo javaPartitionInfo : partitionInfoList) {
+      scalaPartitionInfoSet.add(new TopicAndPartition(_topic, javaPartitionInfo.partition()));
+    }
+    PreferredReplicaLeaderElectionCommand.writePreferredReplicaElectionData(zkUtils, scalaPartitionInfoSet);
+  }
+
+  /**
+   * @return the state of the topic e.g. all the partitions and replicas
+   */
+  private TopicState topicState() {
```
lindong28: getTopicState()?

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -0,0 +1,419 @@
...
+    List<PartitionInfo> partitionInfoList = partitionInfoForMonitoredTopic();
+    Collection<Broker> brokers = scala.collection.JavaConversions.asJavaCollection(_zkUtils.getAllBrokersInCluster());
+    return topicState(partitionInfoList, brokers, _partitionBrokerRatioThreshold);
+  }
+
+
+  private List<PartitionInfo> partitionInfoForMonitoredTopic() {
```
lindong28: `topicState()` (or `getTopicState()`) seems inconsistent with `partitionInfoForMonitoredTopic()`

services/src/main/java/com/linkedin/kmf/services/configs/ProduceServiceConfig.java
```diff
@@ -49,24 +52,17 @@
...
+  public static final String TOPIC_REPLICATION_FACTOR_DOC = "When a topic is created automatically or rebalanced this is the "
       + "replication factor used.";
 
-  public static final String REBALANCE_ENABLED_CONFIG = "produce.topic.rebalanceEnabled";
-  public static final String REBALANCE_ENABLED_DOC = "Periodically move leaders and replica assignments so they are "
-      + "evenly spread amongst brokers.";
-
-  public static final String REBALANCE_PARTITION_MULTIPLE_CONFIG = "produce.topic.rebalancePartitionMultiple";
-  public static final String REBALANCE_PARTITION_MULTIPLE_DOC = "Determines the number of partitions per broker in the ideal case.";
+  public static final String PARTITIONS_PER_BROKER_CONFIG = "topic-management.partitionsPerBroker";
```
lindong28: topic-management.partitionsPerBroker seems inconsistent with topic-management.partitionBrokerRatioThreshold

services/src/main/java/com/linkedin/kmf/services/configs/ProduceServiceConfig.java
```diff
@@ -26,6 +28,7 @@
 
   public static final String TOPIC_CONFIG = CommonServiceConfig.TOPIC_CONFIG;
   public static final String TOPIC_DOC = CommonServiceConfig.TOPIC_DOC;
+  public static final String TOPIC_CONFIG_DEFAULT = "kmf-producer";
```
lindong28: It seems inconsistent to only have a variable for default value of topic.

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -0,0 +1,419 @@
...
+ * monitored topic is sampling all the brokers evenly.
+ */
+public class TopicManagementService implements Service  {
+
+  private static final Logger LOG = LoggerFactory.getLogger(TopicManagementService.class);
+
+  /**
+   * The state of the monitored topic.
+   */
+  static class TopicState {
```
lindong28: After reading `TopicManagementTest.java`, I think we can actually remove TopicState and still be able to do the same unit tests for someBrokerMissingPartition, someBrokerWithoutLeader etc. The idea is to have method like getPreferredLeaders() and use mockit to determine its output.

Removing ToipcState would probably make code more concise. We can discuss it offline.

smccauliff: TopicState is a logically separate concept from the thing that is doing the monitoring and rebalancing.  If it is up to me then I would rather leave it untouched.

lindong28: Yeah I agree.

services/src/main/java/com/linkedin/kmf/services/configs/ProduceServiceConfig.java
smccauliff: Probably CommonServiceConifg should have the TOPIC_CONFIG_DEFAULT.  It's not clear to me why the defaults are not constants already since they can exist in multiple Config classes.

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
smccauliff: So just partitionInfo()?

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
smccauliff: Putting "get" as a prefix for methods is not needed unless the intent is to comply with JavaBeans or have a system of properties like JavaBeans has.  For example the Java collections that were created after Java 1.1 name their methods like "size()" instead of "getSize()".

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
smccauliff: OK.

services/src/main/java/com/linkedin/kmf/services/configs/TopicManagementConfig.java
smccauliff: OK

config/kafka-monitor.properties
```diff
@@ -39,8 +39,11 @@
     "zookeeper.connect": "localhost:2181",
     "bootstrap.servers": "localhost:9092",
     "produce.record.delay.ms": 100,
-    "produce.topic.autoTopicCreationEnabled": true,
-    "produce.topic.autoTopicReplicationFactor" : 1,
+    "produce.topic.topicCreationEnabled": true,
+    "topic-management.replicationFactor" : 1,
+    "topic-management.partitionsPerBroker" : 2,
```
smccauliff: I'm not sure.  Ratios don't need to be natural numbers, but the number of partitions per broker does need to be a natural number.

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
smccauliff: OK

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
smccauliff: _producerPropsOverride can be null.

The new method signature will be initializeProducer() by using _producerPropsOverride and _producerClass.  Yes?

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
smccauliff: I'm not sure.  If we expect the actual value of topic-management.rebalance.interval.ms to be on the order of 10 minutes then it will be an additional 10 minutes before the new partitions are picked up by the thread.

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
smccauliff: OK

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
smccauliff: I've updated the name.    I thought the purpose of starting it inside was so that the configuration would be in the same block in the config file.  Otherwise configuration needs to be duplicated in different blocks.  Is this not correct?

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
smccauliff: I need to look at this more closely before getting back to you.

config/kafka-monitor.properties
smccauliff: Change to:
partitionToBrokerRatio (double, use ceil when creating partitions, must be greater than or eq 0).
partitionToBrokerRatioThreshold (just a name change)

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
smccauliff: We are good with this.

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
smccauliff: The topic management service will again live outside of produce service.  We will duplicate configs. The TMS will need to check if a topic exists and retry later if it does not.  The TMS should be part of the EndToEndTest.

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
smccauliff: OK will reuse this since the configs are duplicated intervals can be configured differently.

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
smccauliff: partitionInfo()

services/src/main/java/com/linkedin/kmf/services/configs/TopicManagementServiceConfig.java
```diff
@@ -0,0 +1,67 @@
...
+
+  static {
+    CONFIG = new ConfigDef()
+      .define(CommonServiceConfig.ZOOKEEPER_CONNECT_CONFIG,
+        ConfigDef.Type.STRING,
+        ConfigDef.Importance.HIGH,
+        CommonServiceConfig.ZOOKEEPER_CONNECT_DOC)
+      .define(CommonServiceConfig.TOPIC_CONFIG,
+        ConfigDef.Type.STRING,
+        "kmf-producer",
```
smccauliff: This is changed in: ProduceServiceConfig, ConsumeServiceConfig, TopicManagementServiceConfig and kafka-monitor.properties

PR #28, opened 2016-10-28
If topic reassignment is in progress then attempt leader election if …

smccauliff:
…needed and do reassignment later.

If we hit the race condition when doing reassignment then catch that ZK exception and try again later.
Assign names to all the threads that ProduceService, ConsumeService and TopicManagementService create,
this way you can distinguish them in dumps, jconsole, jstack and debugger.

### DIFF
--- a/services/src/main/java/com/linkedin/kmf/services/ConsumeService.java
+++ b/services/src/main/java/com/linkedin/kmf/services/ConsumeService.java
@@ -108,7 +108,7 @@ public class ConsumeService implements Service {
           LOG.error(_name + "/ConsumeService failed", e);
         }
       }
-    }, _name + " consume thread");
+    }, _name + " consume-service");
 
     MetricConfig metricConfig = new MetricConfig().samples(60).timeWindow(1000, TimeUnit.MILLISECONDS);
     List<MetricsReporter> reporters = new ArrayList<>();
--- a/services/src/main/java/com/linkedin/kmf/services/ConsumeService.java
+++ b/services/src/main/java/com/linkedin/kmf/services/ConsumeService.java
@@ -108,7 +108,7 @@ public class ConsumeService implements Service {
           LOG.error(_name + "/ConsumeService failed", e);
         }
       }
-    });
+    }, _name + " consume thread");
 
     MetricConfig metricConfig = new MetricConfig().samples(60).timeWindow(1000, TimeUnit.MILLISECONDS);
     List<MetricsReporter> reporters = new ArrayList<>();
--- a/services/src/main/java/com/linkedin/kmf/services/ProduceService.java
+++ b/services/src/main/java/com/linkedin/kmf/services/ProduceService.java
@@ -158,7 +158,7 @@ public class ProduceService implements Service {
   public void start() {
     if (_running.compareAndSet(false, true)) {
       initializeStateForPartitions();
-      _handleNewPartitionsExecutor.scheduleWithFixedDelay(new HandleNewPartitions(), 40_000, 40_000, TimeUnit.MILLISECONDS);
+      _handleNewPartitionsExecutor.scheduleWithFixedDelay(new NewPartitionHandler(), 40_000, 40_000, TimeUnit.MILLISECONDS);
       LOG.info(_name + "/ProduceService started");
     }
   }
@@ -302,7 +302,7 @@ public class ProduceService implements Service {
    * sensors are added for the new partitions.
    */
 
-  private class HandleNewPartitions implements Runnable {
+  private class NewPartitionHandler implements Runnable {
 
     public void run() {
       int currentPartitionCount = Utils.getPartitionNumForTopic(_zkConnect, _topic);
@@ -335,13 +335,13 @@ public class ProduceService implements Service {
 
     private final AtomicInteger _threadId = new AtomicInteger();
     public Thread newThread(Runnable r) {
-      return new Thread(r, _name + " produce-service-produce " + _threadId.getAndIncrement());
+      return new Thread(r, _name + "-produce-service-produce-" + _threadId.getAndIncrement());
     }
   }
 
   private class HandleNewPartitionsThreadFactory implements ThreadFactory {
     public Thread newThread(Runnable r) {
-      return new Thread(r, _name + " produce-service-handle-new-partitions");
+      return new Thread(r, _name + "-produce-service-new-partition-handler");
     }
   }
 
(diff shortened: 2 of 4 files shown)


services/src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -108,7 +108,7 @@ public void run() {
           LOG.error(_name + "/ConsumeService failed", e);
         }
       }
-    });
+    }, _name + " consume thread");
```
lindong28: Should we use `-` instead of space in the thread name?

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
```diff
@@ -330,4 +331,18 @@ public void run() {
...
+
+    private final AtomicInteger _threadId = new AtomicInteger();
+    public Thread newThread(Runnable r) {
+      return new Thread(r, _name + " produce-service, produce thread " + _threadId.getAndIncrement());
+    }
+  }
+
+  private class HandleNewPartitionsThreadFactory implements ThreadFactory {
+    public Thread newThread(Runnable r) {
+      return new Thread(r, _name + " produce-service, handle-new-partitions thread ");
```
lindong28: It seems too verbose to have `,` and `thread` in the thread name. How about `_name + "-produce-service-partition-monitor"`.

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
```diff
@@ -330,4 +331,18 @@ public void run() {
     }
   }
 
+  private class ProduceServiceThreadFactory implements ThreadFactory {
+
+    private final AtomicInteger _threadId = new AtomicInteger();
+    public Thread newThread(Runnable r) {
+      return new Thread(r, _name + " produce-service, produce thread " + _threadId.getAndIncrement());
```
lindong28: It seems too verbose to have , and thread in the thread name. How about `_name + "-produce-service"`?

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -182,7 +184,7 @@ public void run() {
           if (topicState.replicationFactor() == -1) {
             return;
           }
-          if (topicState.someBrokerMissingPartition()) {
+          if (topicState.someBrokerMissingPartition() && !topicAssignmentIsRunning()) {
```
lindong28: Why do we need to wait for topic assignment to complete or even checking if there is ongoing assignment? Even if we do it, we may still see ZkNodeExistsException when we call `reassignPartitions(..)`. We can probably simply do the following:

```
 if (topicState.someBrokerMissingPartition())
  reassignPartitions(topicState._allBrokers, topicState._partitionInfo.size(), topicState.replicationFactor());
```

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -209,6 +214,12 @@ public void run() {
     }
   }
 
+  private class TopicManagementServiceThreadFactory implements ThreadFactory {
+    public Thread newThread(Runnable r) {
+      return new Thread(r, _serviceName + " topic-management-service thread");
```
lindong28: How about _servceName + `-topic-management-service`?

services/src/main/java/com/linkedin/kmf/services/ConsumeService.java
smccauliff: I've removed "thread" from all the thread names.

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
smccauliff: I've removed "thread" from the thread name.  The thread runs the thing that handles new partitions and there are other types of threads that run in the produce service.

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
smccauliff: Why not wait for the topic assignment to complete?  It shouldn't take long for the monitored topic.

lindong28: I agree it makes sense to check if there is ongoing reassignment. But do we still need to call `waitForTopicAssignmentsToComplete()` after this check?

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
lindong28: Can we also replace space with `-`? I think by tradition we don't use space in the any name -- otherwise we may think the words printed before/after the name in the log is part of the name.

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
smccauliff: Why not wait and do it as soon as possible rather checking in another 10 minutes.

lindong28:
@smccauliff Just one more question. Currently in `TopicManagementRunnable.run()` we will print `_serviceName + ": topic is in good state, no rebalance needed."` at INFO level if there is no need for rebalance. I am wondering if we should change this to debug level. Doing this at INFO level will flood log with a lot of useless information.

smccauliff:
Assuming something is printed out every 10 minutes and it's about 80 characters that works out to about 10k bytes per day topic manager.   Even if we were writing to floppy disk drives I would not be concerned about this amount of data.

The feedback I've had from users (not of this software)  is that they want some kind of log message: something that tells them the system is doing something and that the worst thing they can ever see is an empty log.  Since the monitoring thread is not in any kind of I/O path it does not have any kind of activity other than it is periodically woken up to check the state of the monitored topic.  Therefore, this is really the only place to put this kind of log messages.

lindong28:
@smccauliff Yeah the disk space is not an issue at all. I agree that users need to know that the software is running. But I am not convinced that should be in the form of a periodic heartbeat message in the log. Note that the users of Kafka Monitor should be reading metics value from Kafka Monitor (e.g. ProduceAvailability) since that is the sole purpose of running KM. If they are interested in the progress then they can simply read the total produced message count and total consumed message count which are also provided as metrics. If Kafka Monitor is running well then these numbers should be increasing steadily overtime. 

For those users who really want to be sure that TopicManagementService is running by having a period message in the LOG, they can still turn on DEBUG level logging for TopicManagementService in the log4j. This hold for Apache Kafka and probably all Kafka-related service at LinkedIn. I can not find a service that prints this period message at INFO level in the log.

Besides, KM will periodically check if all services are running. If not, then it will print error message in the LOG. If KM stops working properly without printing this as either INFO or ERROR in the log, then developers should look into this and fix the bug.

smccauliff:
If that is the Kafka way...

lindong28:
@smccauliff Thanks for the patch! I just double checked with Becket that we generally don't use INFO level statement to show liveness of services.

PR #29, opened 2016-11-08
Fix two problems preventing shutdown on SIGTERM

smccauliff:
Don't synchronize when waiting for shutdown of topic management service.
KMBaseConsumer implementations should return immediately so that the consume service thread can check for termination.

This was tested by adding a shutdown hook in KafkaMonitor (not part of this pull request) and then sending kill -TERM to the process and seeing the process exit.

### DIFF
--- a/services/src/main/java/com/linkedin/kmf/consumer/KMBaseConsumer.java
+++ b/services/src/main/java/com/linkedin/kmf/consumer/KMBaseConsumer.java
@@ -14,12 +14,6 @@ package com.linkedin.kmf.consumer;
  */
 public interface KMBaseConsumer {
 
-  /**
-   * The implementation should not block for I/O or have a very short poll time so that the calling thread can attempt
-   * to shutdown when the service is stopped.
-   * @return  This may return null if the underlying consumer implementation has not returned any messages.
-   * @throws Exception
-   */
   BaseConsumerRecord receive() throws Exception;
 
   void close();
--- a/services/src/main/java/com/linkedin/kmf/consumer/KMBaseConsumer.java
+++ b/services/src/main/java/com/linkedin/kmf/consumer/KMBaseConsumer.java
@@ -14,6 +14,12 @@ package com.linkedin.kmf.consumer;
  */
 public interface KMBaseConsumer {
 
+  /**
+   * The implementation should not block for I/O or have a very short poll time so that the calling thread can attempt
+   * to shutdown when the service is stopped.
+   * @return  This may return null if the underlying consumer implementation has not returned any messages.
+   * @throws Exception
+   */
   BaseConsumerRecord receive() throws Exception;
 
   void close();
--- a/services/src/main/java/com/linkedin/kmf/consumer/NewConsumer.java
+++ b/services/src/main/java/com/linkedin/kmf/consumer/NewConsumer.java
@@ -30,8 +30,12 @@ public class NewConsumer implements KMBaseConsumer {
 
   @Override
   public BaseConsumerRecord receive() {
-    if (_recordIter == null || !_recordIter.hasNext())
-      _recordIter = _consumer.poll(Long.MAX_VALUE).iterator();
+    if (_recordIter == null || !_recordIter.hasNext()) {
+      _recordIter = _consumer.poll(10).iterator();
+    }
+    if (_recordIter == null || !_recordIter.hasNext()) {
+      return null;
+    }
 
     ConsumerRecord<String, String> record = _recordIter.next();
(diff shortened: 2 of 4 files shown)


services/src/main/java/com/linkedin/kmf/consumer/KMBaseConsumer.java
```diff
@@ -14,6 +14,12 @@
  */
 public interface KMBaseConsumer {
 
+  /**
```
lindong28: Note that restriction is inconvenient for developers who want to plugin their consumer implementation in the Kafka Monitor, because their existing API may only provide a blocking API for receiving messages.

Maybe we don't necessarily need this restriction. How about we simply mark the thread as daemon and don't wait for consumer thread to shutdown when `ConsumeService.stop()` is called?

services/src/main/java/com/linkedin/kmf/consumer/NewConsumer.java
```diff
@@ -30,8 +30,12 @@ public NewConsumer(String topic, Properties consumerProperties) {
 
   @Override
   public BaseConsumerRecord receive() {
-    if (_recordIter == null || !_recordIter.hasNext())
-      _recordIter = _consumer.poll(Long.MAX_VALUE).iterator();
+    if (_recordIter == null || !_recordIter.hasNext()) {
```
lindong28: This is one example where requiring `receive()` to be non-blocking makes implementation more complex.

services/src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -181,7 +183,11 @@ public void start() {
   @Override
   public void stop() {
     if (_running.compareAndSet(true, false)) {
-      _consumer.close();
+      try {
+        _consumer.close();
+      } catch (ConcurrentModificationException cme) {
```
lindong28: Would it be better to cache all exceptions instead of simply the ConcurrentModificationException? The idea is that we want to log all exceptions that may be thrown out of `close()`.

services/src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -33,6 +34,7 @@
 import org.apache.kafka.common.metrics.stats.Total;
 import org.apache.kafka.common.serialization.StringDeserializer;
 import org.apache.kafka.common.utils.SystemTime;
+import org.mortbay.log.Log;
```
lindong28: Do we need this package?

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -391,7 +391,7 @@ public boolean isRunning() {
   }
 
   @Override
-  public synchronized void awaitShutdown() {
+  public void awaitShutdown() {
```
lindong28: Note that start() and close() are also marked with keyward `synchronized`. An althernative, which is the pattern used in other service class, is to not have keyword `synchronized` and make `_running` a `AtomaticBoolean`. Would that be better?

services/src/main/java/com/linkedin/kmf/consumer/KMBaseConsumer.java
smccauliff: OK.  The thread waiting for shutdown will not wait for the consumer thread to exit.  The consumer thread can be made a daemon thread.  This will allow consumers to block indefinitely.

services/src/main/java/com/linkedin/kmf/services/ConsumeService.java
smccauliff: No.  Sloppy use of auto complete on my part.

services/src/main/java/com/linkedin/kmf/services/ConsumeService.java
smccauliff: All exceptions are now caught.

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
lindong28: OK. Let's revisit the choice between `synchronized` and `AtomicBoolean` later. After reviewing TopicManagementService I have got three other comments. Can you take a look?

- Add `LOG.info(_name + "/TopicManagementService started")` in `TopicManagementService.stop()`. Do the similar thing in `TopicManagementService.start()`
- Set _running to false in `TopicManagementService.stop()`
- Only do `_executor.awaitTermination(...)` in `TopicManagementService.awaitShutdown()`.

PR #33, opened 2016-11-08
ProduceService and ConsumeService should log exception from produce() and consume() at WARN level

lindong28:


smccauliff:
Looks good to me.

lindong28:
@smccauliff Thanks. Can you take another look? I removed check for `_running` in `TopicManagementService.awaitShutdown()`

smccauliff:
Yeah.  The TopicManagementService running means that someone no longer needs to wait for it to stop running.

PR #34, opened 2016-12-01
Fix potential use of two separate zookeepers for the same producer

WarrenGreen:
Currently a zookeeper config in ProduceService can be used to populate existingPartitionCount or create a new topic while the override zookeeper config will be passed to the actual producer. This allows only one zookeeper config to be specified.

### DIFF
--- a/services/src/main/java/com/linkedin/kmf/services/ProduceService.java
+++ b/services/src/main/java/com/linkedin/kmf/services/ProduceService.java
@@ -97,15 +97,6 @@ public class ProduceService implements Service {
       }
     }
 
-    _producerPropsOverride = props.containsKey(ProduceServiceConfig.PRODUCER_PROPS_CONFIG) ?
-      (Map) props.get(ProduceServiceConfig.PRODUCER_PROPS_CONFIG) : new HashMap<>();
-
-    for (String property: NONOVERRIDABLE_PROPERTIES) {
-      if (_producerPropsOverride.containsKey(property)) {
-        throw new ConfigException("Override must not contain bootsrap or zookeeper config.");
-      }
-    }
-
     double partitionsToBrokersRatio = config.getDouble(CommonServiceConfig.PARTITIONS_TO_BROKER_RATO_CONFIG);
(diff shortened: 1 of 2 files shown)


services/src/main/java/com/linkedin/kmf/services/ProduceService.java
```diff
@@ -87,6 +89,15 @@ public ProduceService(Map<String, Object> props, String name) throws Exception {
     _running = new AtomicBoolean(false);
     _nextIndexPerPartition = new ConcurrentHashMap<>();
 
+    _producerPropsOverride = props.containsKey(ProduceServiceConfig.PRODUCER_PROPS_CONFIG) ?
+      (Map) props.get(ProduceServiceConfig.PRODUCER_PROPS_CONFIG) : new HashMap<>();
+
+    for (String property: NONOVERRIDABLE_PROPERTIES) {
+      if (_producerPropsOverride.containsKey(property)) {
+        throw new ConfigException("Override must not contain bootsrap or zookeeper config.");
```
smccauliff: bootsrap -> bootstrap

lindong28:
Thanks for the patch. Can you make the same change for ConsumeService a s well?

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
```diff
@@ -86,6 +88,14 @@ public ProduceService(Map<String, Object> props, String name) throws Exception {
     _partitionNum = new AtomicInteger(0);
     _running = new AtomicBoolean(false);
     _nextIndexPerPartition = new ConcurrentHashMap<>();
+    _producerPropsOverride = props.containsKey(ProduceServiceConfig.PRODUCER_PROPS_CONFIG) ?
+      (Map) props.get(ProduceServiceConfig.PRODUCER_PROPS_CONFIG) : new HashMap<>();
+
+    for (String property: NONOVERRIDABLE_PROPERTIES) {
+      if (_producerPropsOverride.containsKey(property)) {
+        throw new ConfigException("Override must not contain bootstrap or zookeeper config.");
```
lindong28: Should this message be `"Property " + property " should not be specified inside " + ProduceServiceConfig.PRODUCER_PROPS_CONFIG`?

Same for the error message in ConsumeService

PR #36, opened 2016-12-01
Add topic creation to ConsumeService and minor text formatting

WarrenGreen:
- Allow for topic creation to work when using separate produce/consume clusters by adding topic creation in the ConsumeService
- Fix minor text formatting in ProduceService

### DIFF
--- a/config/kafka-monitor.properties
+++ b/config/kafka-monitor.properties
@@ -39,7 +39,7 @@
     "zookeeper.connect": "localhost:2181",
     "bootstrap.servers": "localhost:9092",
     "produce.record.delay.ms": 100,
-    "produce.topic.topicCreationEnabled": true,
+    "topic-management.topicCreationEnabled": true,
     "topic-management.replicationFactor" : 1,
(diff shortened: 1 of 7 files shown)


lindong28:
Can you explain why should ConsumeService create topic? In general the application who produce data will create topic.

WarrenGreen:
When monitoring a mirror maker across two clusters, it will currently only create a topic on the producer cluster and will not on the consumer cluster. With this change, it will create the topic as needed on both clusters. If there is only one cluster that both the producer/consumer use then one of them will successfully create the topic and the other will return the # of partitions that were created for the topic.

smccauliff:
I'm confused about this can you show me on the white board?

lindong28:
But adding this in ConsumeService wouldn't solve the problem where we monitor a pipline of three cluster. The topic in the middle middle also needs to have proper number of partitions and these partitions need to be distributed across broker properly, which is the functionality of the TopicManagementService. Instead of asking ConsumeService to create topic, maybe we should simply instantiate one TopicManagementService per cluster.

PR #40, opened 2016-12-13
Move topic creation from produce service to Topic Management Service.

WarrenGreen:
Topic creation is now a responsibility of the TopicManagementService and the ProducerService will non-blocking wait to produce until the topic is created.

### DIFF
--- a/src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
+++ b/src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
@@ -204,7 +204,7 @@ public class SingleClusterMonitor implements App {
       .type(Boolean.class)
       .metavar("AUTO_TOPIC_CREATION_ENABLED")
       .dest("autoTopicCreationEnabled")
-      .help(ProduceServiceConfig.TOPIC_CREATION_ENABLED_DOC);
+      .help(TopicManagementServiceConfig.TOPIC_CREATION_ENABLED_DOC);
 
     parser.addArgument("--topic-rebalance-interval-ms")
       .action(store())
@@ -244,7 +244,7 @@ public class SingleClusterMonitor implements App {
     if (res.getString("producerConfig") != null)
       props.put(ProduceServiceConfig.PRODUCER_PROPS_CONFIG, Utils.loadProps(res.getString("producerConfig")));
     if (res.getBoolean("autoTopicCreationEnabled") != null)
-      props.put(ProduceServiceConfig.TOPIC_CREATION_ENABLED_CONFIG, res.getBoolean("autoTopicCreationEnabled"));
+      props.put(TopicManagementServiceConfig.TOPIC_CREATION_ENABLED_CONFIG, res.getBoolean("autoTopicCreationEnabled"));
     if (res.getInt("rebalanceMs") != null)
       props.put(TopicManagementServiceConfig.REBALANCE_INTERVAL_MS_CONFIG, res.getInt("rebalanceMs"));
 
@@ -260,8 +260,8 @@ public class SingleClusterMonitor implements App {
     if (res.getString("latencyPercentileGranularityMs") != null)
       props.put(ConsumeServiceConfig.LATENCY_PERCENTILE_GRANULARITY_MS_CONFIG, res.getString("latencyPercentileGranularityMs"));
 
-    SingleClusterMonitor test = new SingleClusterMonitor(props, "end-to-end");
-    test.start();
+    SingleClusterMonitor app = new SingleClusterMonitor(props, "end-to-end");
+    app.start();
 
     // metrics export service config
     props = new HashMap<>();
@@ -289,6 +289,6 @@ public class SingleClusterMonitor implements App {
     JettyService jettyService = new JettyService(new HashMap<String, Object>(), "end-to-end");
     jettyService.start();
 
-    test.awaitShutdown();
+    app.awaitShutdown();
   }
 }
--- a/config/kafka-monitor.properties
+++ b/config/kafka-monitor.properties
@@ -39,7 +39,7 @@
     "zookeeper.connect": "localhost:2181",
     "bootstrap.servers": "localhost:9092",
     "produce.record.delay.ms": 100,
-    "produce.topic.topicCreationEnabled": true,
+    "topic-management.topicCreationEnabled": true,
     "topic-management.replicationFactor" : 1,
     "topic-management.partitionsToBrokerRatio" : 2.0,
     "topic-management.partitionsToBrokersRatioThreshold" : 1.5,
--- a/src/main/java/com/linkedin/kmf/common/Utils.java
+++ b/src/main/java/com/linkedin/kmf/common/Utils.java
@@ -9,6 +9,7 @@
  */
 package com.linkedin.kmf.common;
 
+import java.util.NoSuchElementException;
 import java.util.Properties;
 import kafka.admin.RackAwareMode;
 import kafka.server.KafkaConfig;
@@ -50,6 +51,8 @@ public class Utils {
     try {
       Seq<String> topics = scala.collection.JavaConversions.asScalaBuffer(Arrays.asList(topic));
       return zkUtils.getPartitionsForTopics(topics).apply(topic).size();
+    } catch (NoSuchElementException e) {
+      return 0;
     } finally {
       zkUtils.close();
     }
--- a/src/main/java/com/linkedin/kmf/services/ProduceService.java
+++ b/src/main/java/com/linkedin/kmf/services/ProduceService.java
@@ -10,15 +10,24 @@
 package com.linkedin.kmf.services;
 
 import com.linkedin.kmf.common.Utils;
-import com.linkedin.kmf.services.configs.ProduceServiceConfig;
-import com.linkedin.kmf.producer.KMBaseProducer;
 import com.linkedin.kmf.producer.BaseProducerRecord;
+import com.linkedin.kmf.producer.KMBaseProducer;
 import com.linkedin.kmf.producer.NewProducer;
+import com.linkedin.kmf.services.configs.ProduceServiceConfig;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
-import com.linkedin.kmf.topicfactory.TopicFactory;
+import java.util.concurrent.atomic.AtomicLong;
 import org.apache.kafka.clients.producer.ProducerConfig;
 import org.apache.kafka.clients.producer.RecordMetadata;
 import org.apache.kafka.common.MetricName;
@@ -34,16 +43,6 @@ import org.apache.kafka.common.metrics.stats.Total;
 import org.apache.kafka.common.utils.SystemTime;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Properties;
-import java.util.concurrent.Executors;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicLong;
 
 
 public class ProduceService implements Service {
@@ -90,8 +89,8 @@ public class ProduceService implements Service {
     _partitionNum = new AtomicInteger(0);
     _running = new AtomicBoolean(false);
     _nextIndexPerPartition = new ConcurrentHashMap<>();
-    _producerPropsOverride = props.containsKey(ProduceServiceConfig.PRODUCER_PROPS_CONFIG) ?
-      (Map) props.get(ProduceServiceConfig.PRODUCER_PROPS_CONFIG) : new HashMap<>();
+    _producerPropsOverride = props.containsKey(ProduceServiceConfig.PRODUCER_PROPS_CONFIG)
+      ? (Map) props.get(ProduceServiceConfig.PRODUCER_PROPS_CONFIG) : new HashMap<>();
 
     for (String property: NONOVERRIDABLE_PROPERTIES) {
       if (_producerPropsOverride.containsKey(property)) {
@@ -297,8 +296,10 @@ public class ProduceService implements Service {
 
     public void run() {
       int currentPartitionCount = Utils.getPartitionNumForTopic(_zkConnect, _topic);
-      if (currentPartitionCount == _partitionNum.get()) {
+      if (currentPartitionCount <= 0) {
         LOG.info(_name + ": Topic named " + _topic + " does not exist.");
+        return;
+      } else if (currentPartitionCount == _partitionNum.get()) {
         return;
       }
       LOG.info(_name + ": Detected new partitions to monitor.");
--- a/src/main/java/com/linkedin/kmf/services/ProduceService.java
+++ b/src/main/java/com/linkedin/kmf/services/ProduceService.java
@@ -100,10 +100,7 @@ public class ProduceService implements Service {
       }
     }
 
-    double partitionsToBrokersRatio = config.getDouble(CommonServiceConfig.PARTITIONS_TO_BROKER_RATO_CONFIG);
-    int topicReplicationFactor = config.getInt(ProduceServiceConfig.TOPIC_REPLICATION_FACTOR_CONFIG);
     int existingPartitionCount = Utils.getPartitionNumForTopic(_zkConnect, _topic);
-
     if (existingPartitionCount <= 0) {
       if (config.getBoolean(ProduceServiceConfig.TOPIC_CREATION_ENABLED_CONFIG)) {
         TopicFactory topicFactory =
@@ -113,9 +110,8 @@ public class ProduceService implements Service {
       } else {
         throw new RuntimeException("Can not find valid partition number for topic " + _topic +
             ". Please verify that the topic \"" + _topic + "\" has been created. Ideally the partition number should be" +
-            " a multiple of number" +
-            " of brokers in the cluster.  Or else configure " + ProduceServiceConfig.TOPIC_CREATION_ENABLED_CONFIG +
-            " to be true.");
+            " a multiple of number of brokers in the cluster.  Or else configure " +
+            CommonServiceConfig.TOPIC_CREATION_ENABLED_CONFIG + " to be true.");
       }
     } else {
       _partitionNum.set(existingPartitionCount);
--- a/src/main/java/com/linkedin/kmf/services/ProduceService.java
+++ b/src/main/java/com/linkedin/kmf/services/ProduceService.java
@@ -99,15 +99,7 @@ public class ProduceService implements Service {
       }
     }
 
-    int existingPartitionCount = Utils.getPartitionNumForTopic(_zkConnect, _topic);
-    int topicPollDelay = config.getInt(ProduceServiceConfig.TOPIC_POLL_DELAY_CONFIG);
-    while (existingPartitionCount <= 0) {
-      LOG.info(_name + ": Topic named " +  _topic + " has not been created. Checking again in " + topicPollDelay + "ms.");
-      Thread.sleep(topicPollDelay);
-      existingPartitionCount = Utils.getPartitionNumForTopic(_zkConnect, _topic);
-    }
-
-    _partitionNum.set(existingPartitionCount);
+    _partitionNum.set(Utils.getPartitionNumForTopic(_zkConnect, _topic));
 
     if (producerClass.equals(NewProducer.class.getCanonicalName()) || producerClass.equals(NewProducer.class.getSimpleName())) {
       _producerClassName = NewProducer.class.getCanonicalName();
@@ -306,6 +298,7 @@ public class ProduceService implements Service {
     public void run() {
       int currentPartitionCount = Utils.getPartitionNumForTopic(_zkConnect, _topic);
       if (currentPartitionCount == _partitionNum.get()) {
+        LOG.info(_name + ": Topic named " + _topic + " does not exist.");
         return;
       }
       LOG.info(_name + ": Detected new partitions to monitor.");
--- a/src/main/java/com/linkedin/kmf/services/ProduceService.java
+++ b/src/main/java/com/linkedin/kmf/services/ProduceService.java
@@ -10,7 +10,6 @@
 package com.linkedin.kmf.services;
 
 import com.linkedin.kmf.common.Utils;
-import com.linkedin.kmf.services.configs.CommonServiceConfig;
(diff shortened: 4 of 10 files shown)


services/src/main/java/com/linkedin/kmf/services/ProduceService.java
```diff
@@ -97,21 +97,20 @@ public ProduceService(Map<String, Object> props, String name) throws Exception {
       }
     }
 
-    double partitionsToBrokersRatio = config.getDouble(CommonServiceConfig.PARTITIONS_TO_BROKER_RATO_CONFIG);
-    int topicReplicationFactor = config.getInt(ProduceServiceConfig.TOPIC_REPLICATION_FACTOR_CONFIG);
+    double partitionsToBrokersRatio = 1;
```
smccauliff: Why not just remove these local variables entirely?

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -113,6 +116,14 @@ boolean someBrokerWithoutLeader() {
     }
 
     /**
+     *
+     * @return true if the current topic replication factor is less than the expected ratio
+     */
+    boolean insufficientReplicationFactor() {
+      return replicationFactor() < _topicReplicationFactor;
```
smccauliff: _topicReplicationFactor might be better called _configuredReplicationFactor.  replicationFactor() can then be renamed something like currentReplicationFactor().

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
WarrenGreen: Good idea.

services/src/main/java/com/linkedin/kmf/services/configs/CommonServiceConfig.java
```diff
@@ -27,4 +27,13 @@
     + " created or rebalanced.  ceil(nBrokers * partitionsToBrokerRatio) is used to determine the actual number of "
     + "partitions when partitions are added or removed.";
 
+  public static final String TOPIC_REPLICATION_FACTOR_CONFIG = "topic-management.replicationFactor";
```
lindong28: Why is this config in `CommonServiceConfig` instead of `TopicManagementServiceConfig`?

Can you help correct typo in the PARTITIONS_TO_BROKER_RATO_CONFIG and PARTITIONS_TO_BROKER_RATIO_DOC? Also, that config is also used only in TopicManagementService. Can you move it as well?

services/src/main/java/com/linkedin/kmf/services/configs/CommonServiceConfig.java
```diff
@@ -27,4 +27,13 @@
...
+  public static final String TOPIC_REPLICATION_FACTOR_DOC = "When a topic is created automatically this is the "
+      + "replication factor used.";
+
+  public static final String TOPIC_CREATION_ENABLED_CONFIG = "topic-management.topicCreationEnabled";
```
lindong28: Given that we have already release kafka-monitor as open source, let's no make this backward incompatible change by changing the config name, unless it is absolutely necessary for new use-case.

Besides, if it is a common config used by ProduceService and TopicManagementService, this config name should not have `topic-management` as prefix.

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -62,20 +63,22 @@
     final Set<Integer> _electedLeaders;
     final Collection<Broker> _allBrokers;
     final double _partitionToBrokerRatioThreshold;
+    final int _topicReplicationFactor;
```
lindong28: `_topicReplicationFactor` should not be part of the `TopicState` because it is statically configured and will have the same value for all `TopicState` instances. It should be put in the same place as `_topicCreationEnabled`.

Besides, why rename it from `_topicReplicationFactor` to `_configuredReplicationFactor`?

services/src/main/java/com/linkedin/kmf/services/configs/TopicManagementServiceConfig.java
```diff
@@ -58,7 +58,18 @@
               2.0,
               atLeast(1),
               ConfigDef.Importance.LOW,
-              CommonServiceConfig.PARTITIONS_TO_BROKER_RATIO_DOC);
+              CommonServiceConfig.PARTITIONS_TO_BROKER_RATIO_DOC)
+      .define(CommonServiceConfig.TOPIC_CREATION_ENABLED_CONFIG,
```
lindong28: Can you correct the indentation?

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -162,14 +173,18 @@ public void run() {
...
         }
-        if (topicState.someBrokerWithoutLeader() || topicState.someBrokerMissingPartition() || topicState.insufficientPartitions()) {
+        if (topicState.someBrokerWithoutLeader() || topicState.someBrokerMissingPartition() || topicState.insufficientPartitions()
+            || topicState.insufficientReplicationFactor()) {
 
           LOG.info(_serviceName + ": topic rebalance started.");
           LOG.debug(_serviceName + ": broker count " + topicState._allBrokers.size() + " partitionCount " + topicState._partitionInfo.size() + ".");
 
-          if (topicState.replicationFactor() == -1) {
+          if (topicState.currentReplicationFactor() == -1) {
```
lindong28: Here the `topicState.currentReplicationFactor` is called 3 times and `topicState.insufficientReplicationFactor` is called 1 time. This makes code unnecessarily messy. Plus this is beyond the scope of this patch which says "Move topic creation from produce service to Topic Management Service".

At first glance I am not really convinced that we should have this check, and whether this is the best way to do this check. For things that is not straightforward, they should probably be put in their own pull request with proper explanation of why they are needed. Maybe you can remove it from this patch?

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -229,9 +254,24 @@ public TopicManagementService(Map<String, Object> props, String serviceName) {
     _topic = config.getString(CommonServiceConfig.TOPIC_CONFIG);
     _partitionsToBrokerRatio = config.getDouble(CommonServiceConfig.PARTITIONS_TO_BROKER_RATO_CONFIG);
     _scheduleIntervalMs = config.getInt(TopicManagementServiceConfig.REBALANCE_INTERVAL_MS_CONFIG);
-    String zkUrl = config.getString(CommonServiceConfig.ZOOKEEPER_CONNECT_CONFIG);
-    _zkUtils = ZkUtils.apply(zkUrl, ZK_SESSION_TIMEOUT_MS, ZK_CONNECTION_TIMEOUT_MS, JaasUtils.isZkSecurityEnabled());
+    _zkConnect = config.getString(CommonServiceConfig.ZOOKEEPER_CONNECT_CONFIG);
```
lindong28: Why do you change this from local variable to class variable? It is not used anywhere else in the class.

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -229,9 +254,24 @@ public TopicManagementService(Map<String, Object> props, String serviceName) {
...
+    _zkUtils = ZkUtils.apply(_zkConnect, ZK_SESSION_TIMEOUT_MS, ZK_CONNECTION_TIMEOUT_MS, JaasUtils.isZkSecurityEnabled());
     _executor = Executors.newSingleThreadScheduledExecutor(new TopicManagementServiceThreadFactory());
+    _configuredReplicationFactor = config.getInt(CommonServiceConfig.TOPIC_REPLICATION_FACTOR_CONFIG);
+    _topicCreationEnabled = config.getBoolean(CommonServiceConfig.TOPIC_CREATION_ENABLED_CONFIG);
+
+    int existingPartitionCount = Utils.getPartitionNumForTopic(_zkConnect, _topic);
```
lindong28: It is not good to duplicate code in the project. Maybe it can be moved to Utils.java.

services/src/main/java/com/linkedin/kmf/services/configs/ProduceServiceConfig.java
```diff
@@ -111,18 +96,11 @@
...
-                                    atLeast(1),
-                                    ConfigDef.Importance.LOW,
-                                    TOPIC_REPLICATION_FACTOR_DOC)
-                            .define(CommonServiceConfig.PARTITIONS_TO_BROKER_RATO_CONFIG,
-                                    ConfigDef.Type.DOUBLE,
-                                    2.0,
-                                    atLeast(1.0),
-                                    ConfigDef.Importance.LOW,
-                                    CommonServiceConfig.PARTITIONS_TO_BROKER_RATIO_DOC);
+                            .define(CommonServiceConfig.TOPIC_CREATION_ENABLED_CONFIG,
```
lindong28: Can you correct the indentation?

services/src/main/java/com/linkedin/kmf/services/configs/CommonServiceConfig.java
WarrenGreen: Done

services/src/main/java/com/linkedin/kmf/services/configs/CommonServiceConfig.java
WarrenGreen: Discussed offline

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
WarrenGreen: Discussed offline

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
WarrenGreen: Done.

The name change was an attempt to show the difference between the topics current replication factor and the replication factor passed in as a config.

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
WarrenGreen: Discussed Offline.

services/src/main/java/com/linkedin/kmf/services/configs/TopicManagementServiceConfig.java
WarrenGreen: Done

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
```diff
@@ -97,26 +97,16 @@ public ProduceService(Map<String, Object> props, String name) throws Exception {
...
-        throw new RuntimeException("Can not find valid partition number for topic " + _topic +
-            ". Please verify that the topic \"" + _topic + "\" has been created. Ideally the partition number should be" +
-            " a multiple of number" +
-            " of brokers in the cluster.  Or else configure " + ProduceServiceConfig.TOPIC_CREATION_ENABLED_CONFIG +
-            " to be true.");
-      }
-    } else {
-      _partitionNum.set(existingPartitionCount);
+    int topicPollDelay = config.getInt(ProduceServiceConfig.TOPIC_POLL_DELAY_CONFIG);
+    while (existingPartitionCount <= 0) {
```
lindong28: The sleep and wait logic should be moved from constructor to start().

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -69,7 +70,7 @@
      * @param partitionToBrokerRatioThreshold the expected partition to broker ratio
      */
     TopicState(Set<Integer> electedLeaders, Set<Integer> preferredLeaders, Collection<Broker> allBrokers,
-      List<PartitionInfo> partitionInfo, double partitionToBrokerRatioThreshold) {
+      List<PartitionInfo> partitionInfo, double partitionToBrokerRatioThreshold, int topicReplicationFactor) {
```
lindong28: Why do you need `topicReplicationFactor` in `TopicState`?

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -162,14 +163,25 @@ public void run() {
       try {
         TopicState topicState = topicState();
         if (topicState == null) {
+          if(_topicCreationEnabled) {
```
lindong28: You probably needs to think about how to make the logic in `run()` clean and easy to read.

For example, `topicState.currentReplicationFactor()` is called 3 times. I am not if they are all needed. If topic auto create is not enabled and topic is not created, it doesn't make sense to simply do `return` in `run()`.

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -216,22 +228,39 @@ public Thread newThread(Runnable r) {
...
+    _zkConnect = config.getString(CommonServiceConfig.ZOOKEEPER_CONNECT_CONFIG);
+    _zkUtils = ZkUtils.apply(_zkConnect, ZK_SESSION_TIMEOUT_MS, ZK_CONNECTION_TIMEOUT_MS, JaasUtils.isZkSecurityEnabled());
     _executor = Executors.newSingleThreadScheduledExecutor(new TopicManagementServiceThreadFactory());
+    _topicReplicationFactor = config.getInt(TopicManagementServiceConfig.TOPIC_REPLICATION_FACTOR_CONFIG);
+    _topicCreationEnabled = config.getBoolean(TopicManagementServiceConfig.TOPIC_CREATION_ENABLED_CONFIG);
+
+    int existingPartitionCount = Utils.getPartitionNumForTopic(_zkConnect, _topic);
+    if (existingPartitionCount <= 0) {
+      if (_topicCreationEnabled) {
+        Utils.createMonitoringTopicIfNotExists(_zkConnect, _topic, _topicReplicationFactor, _partitionsToBrokerRatio);
```
lindong28: Do you actually need two places to have this logic of topic creation? Why should this be put?

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -216,22 +228,39 @@ public Thread newThread(Runnable r) {
...
     _serviceName = serviceName;
     TopicManagementServiceConfig config = new TopicManagementServiceConfig(props);
     _partitionToBrokerRatioThreshold = config.getDouble(TopicManagementServiceConfig.PARTITIONS_TO_BROKER_RATIO_THRESHOLD);
     _topic = config.getString(CommonServiceConfig.TOPIC_CONFIG);
-    _partitionsToBrokerRatio = config.getDouble(CommonServiceConfig.PARTITIONS_TO_BROKER_RATO_CONFIG);
+    _partitionsToBrokerRatio = config.getDouble(TopicManagementServiceConfig.PARTITIONS_TO_BROKER_RATIO_CONFIG);
     _scheduleIntervalMs = config.getInt(TopicManagementServiceConfig.REBALANCE_INTERVAL_MS_CONFIG);
-    String zkUrl = config.getString(CommonServiceConfig.ZOOKEEPER_CONNECT_CONFIG);
-    _zkUtils = ZkUtils.apply(zkUrl, ZK_SESSION_TIMEOUT_MS, ZK_CONNECTION_TIMEOUT_MS, JaasUtils.isZkSecurityEnabled());
+    _zkConnect = config.getString(CommonServiceConfig.ZOOKEEPER_CONNECT_CONFIG);
```
lindong28: Do you need `_zkConnect` be class private variable?

services/src/main/java/com/linkedin/kmf/services/configs/ProduceServiceConfig.java
```diff
@@ -29,6 +29,9 @@
   public static final String TOPIC_CONFIG = CommonServiceConfig.TOPIC_CONFIG;
   public static final String TOPIC_DOC = CommonServiceConfig.TOPIC_DOC;
 
+  public static final String TOPIC_POLL_DELAY_CONFIG = "produce.topic.poll.delay";
```
lindong28: Do we need this config? In which case do users need to explicitly set value for this config?

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -118,7 +119,7 @@ boolean someBrokerWithoutLeader() {
      * @return -1 if partitions do not all have the same replication factor possibly indicating that the topic is
      * undergoing maintenance
      */
-    int replicationFactor() {
+    int currentReplicationFactor() {
```
lindong28: I am not sure if we should make this `currentReplicationFactor`. By definition any state of `TopicState` should be current, right?

smccauliff: There are places where currentReplicationFactor() gets compared with _replicationFactor which makes the code hard to read.

lindong28: Would it be better to change _replicationFactor to `_configuredReplicationFactor`, and use `replicationFactor()` instead of `currentReplicationFactor()`? I don't think `current` tells much useful information. It is not `current` vs. `configured` or `current` vs. `previous`. It is `actual` vs. `configured`.

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
lindong28: Neither start() nor the constructor should block waiting for topic to be created.

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
smccauliff: As long as TopicState.replicationFactor() is distinct from _replicationFactor which one is renamed is not important.

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
smccauliff: At some point we need to document the contract implied by com.linkedin.kmf.services.Service.

lindong28: I agree.

WarrenGreen: Discussed offline.

NewPartitionHandler already performs this check and manages the producer thread pool.

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
WarrenGreen: Changed to replicationFactor() and _configuredReplicationFactor

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
WarrenGreen: It's used to create a new topic in TopicManagementRunnable

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
WarrenGreen: Apologies. Must have been reintroduced in a merge.

services/src/main/java/com/linkedin/kmf/services/configs/ProduceServiceConfig.java
WarrenGreen: New design doesn't rely on polling in this manner. Removed.

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -162,30 +163,46 @@ public void run() {
...
+            topicFactory.createTopicIfNotExist(_zkConnect, _topic, _configuredTopicReplicationFactor,
+              _partitionsToBrokerRatio, new Properties());
+            topicState = topicState();
+          } else {
+            throw new RuntimeException(_serviceName + ": topic, " + _topic + ", does not exist and topic creation is not enabled.");
+          }
+        }
+
+        int currentReplicationFactor = topicState.replicationFactor();
+        if (currentReplicationFactor == -1) {
```
lindong28: Would it be better to do this?

```
if (currentReplicationFactor < 0) {
  LOG.info(...)
} else if (currentReplicationFactor != _configuredTopicReplicationFactor) {
  throw new RuntimeException();
}
```

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -162,30 +163,46 @@ public void run() {
...
-            LOG.info(_serviceName + " can't determine replication factor of monitored topic " + _topic + ".  Will try rebalance later.");
-            return;
-          }
           if (topicState.insufficientPartitions()) {
             int idealPartitionCount = (int) Math.ceil(_partitionsToBrokerRatio * topicState._allBrokers.size());
             int addPartitionCount = idealPartitionCount - topicState._partitionInfo.size();
             LOG.info(_serviceName + ": adding " + addPartitionCount + " partitions.");
             AdminUtils.addPartitions(_zkUtils, _topic, idealPartitionCount, null, false, RackAwareMode.Enforced$.MODULE$);
             topicState = topicState();
+            currentReplicationFactor = topicState.replicationFactor();
```
lindong28: Do we need to update and check replicationFactor() again? I understand it is the existing logic. But it seems that we don't need it. replicationFactor() will be -1 only if this topic is being changed by another user. But since we already checked it at the beginning of this method, the time window of this happening is already small and we can not eliminate this problem even if we check it again here.

WarrenGreen: @smccauliff would know the answer to this better than I

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
WarrenGreen: Yeah, that's a much cleaner solution

services/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
smccauliff: run() has become a very long method and should probably be refactored into smaller methods like : createTopic and rebalance(), but that can be some other time as this PR has been out here for sometime and is already very out of sync with master.

lindong28: @smccauliff Yeah, I agree that this check can help catch the case if this is concurrent change by another tool between check at line 176 and this check at line 196. But it is equally likely that the change by another tool is made after line 196 which may mess up TopicManagmentService anyway. Thus I think the increased protection is not worth the additional code.

src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
```diff
@@ -260,8 +260,8 @@ public static void main(String[] args) throws Exception {
     if (res.getString("latencyPercentileGranularityMs") != null)
       props.put(ConsumeServiceConfig.LATENCY_PERCENTILE_GRANULARITY_MS_CONFIG, res.getString("latencyPercentileGranularityMs"));
 
-    SingleClusterMonitor test = new SingleClusterMonitor(props, "end-to-end");
-    test.start();
+    SingleClusterMonitor app = new SingleClusterMonitor(props, "end-to-end");
```
WarrenGreen: @lindong28 Although kind of unrelated to this commit. I feel this name is better given the new hierarchy

lindong28: @WarrenGreen Thanks for the patch! I will review the patch later today.

lindong28: @WarrenGreen Sorry for the delay... Reviewing it now.

lindong28:
Thanks much for the patch! The patch has largely greatly the topic management logic.

PR #42, opened 2016-12-20
Add a test framework for monitoring across mirror maker pipelines.

WarrenGreen:


tests/src/main/java/com/linkedin/kmf/tests/BasicPipelineTest.java
```diff
@@ -0,0 +1,293 @@
...
+ * integer in the message String, consumes messages, and keeps track of number of lost messages, duplicate messages, e2e
+ * latency, throughput, etc. As well, BasicPipeLineTest creates a topic management service for every kafka and zookeeper
+ * broker in the pipeline.
+ *
+ * BasicPipelineTest test exports these metrics via JMX. It also periodically report metrics if DEBUG level logging
+ * is enabled. This information can be used by other application to trigger alert when kafka brokers fail. It also
+ * allows users to track pipeline performance, e.g. latency and throughput.
+ */
+
+public class BasicPipelineTest implements Test {
```
lindong28: The first thing we need to think about BasicPipelineTest is its interface, i.e. how do you want users to specify its configuration. Note that there are two different ways to specify configuration, i.e. command line and config file. Let's work on the config file first since the command line approach will depend on the config file approach. In your current patch, you have asked user to specify `consume.props` for consume service and similarly `produce.props` for produce  service. An alternative is to ask user to provide bootstrap.server and zookeeper.url of source cluster and destination cluster directly, maybe with config name such as source.zookeeper.url which will be specific to BasicPipelineTest, and construct properties map for ProduceService and ConsumeService based on the user-provided config.

I personally think the 2nd approach is better because user doesn't have to know how to config ProduceService and ConsumeService when configuring BasicPipelineTest, 2) it can potentially make the config more intuitive with dedicated config name (e.g. source.zookeeper.url), and 3) less config level. It means that we need to have an config class for each test, similar to what we already have for each service. We can discuss this design when we meet next time.

Also, suppose you go with current approach of having consume.props, I don't think it should be in the ConsumeServiceConfig. For example, let's say in the future we have a test which needs to create to ConsumeService for both source and destination cluster. How are you going to specify the config for this test? Instead of having a config with name `consume.props`, you probably need `source.consume.props` and `destination.consume.props`. It means that this config name should be specific to each test instead of being part of the service.

lindong28: Actually I am not very strong which approach is better before comparing how the properties file would look like with a concrete example. Can you provide an example properties which you used to test BasicPipelineTest?

tests/src/main/java/com/linkedin/kmf/tests/BasicPipelineTest.java
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    zookeeperUrlsList.add(getString(props, ProduceServiceConfig.ZOOKEEPER_CONNECT_CONFIG));
+    zookeeperUrlsList.add(getString(props, ConsumeServiceConfig.ZOOKEEPER_CONNECT_CONFIG));
+
+    Set<String> zookeeperUrls = new HashSet<>();
+    zookeeperUrls.addAll(zookeeperUrlsList);
+    for (String zookeeper: zookeeperUrls) {
+      topicManagementConfig.put(ConsumeServiceConfig.ZOOKEEPER_CONNECT_CONFIG, zookeeper);
+      _topicManagementServices.add(new TopicManagementService(topicManagementConfig, name));
```
lindong28: We probably need to use different names for those services so that we can easily differentiate between them in the log.

tests/src/main/java/com/linkedin/kmf/tests/BasicPipelineTest.java
```diff
@@ -0,0 +1,293 @@
...
+    List<String> zookeeperUrlsList = getList(topicManagementConfig, ConsumeServiceConfig.ZOOKEEPER_CONNECT_CONFIG);
+    if(zookeeperUrlsList == null) {
+      throw new RuntimeException(_name + ": Must specify at least one zookeeper connection.");
+    }
+
+    zookeeperUrlsList.add(getString(props, ProduceServiceConfig.ZOOKEEPER_CONNECT_CONFIG));
+    zookeeperUrlsList.add(getString(props, ConsumeServiceConfig.ZOOKEEPER_CONNECT_CONFIG));
+
+    Set<String> zookeeperUrls = new HashSet<>();
+    zookeeperUrls.addAll(zookeeperUrlsList);
```
lindong28: Instead of using a hashset to remove duplicates, it is probably better to just throw exception if user has provided a cluster more than once in this config.

tests/src/main/java/com/linkedin/kmf/tests/BasicPipelineTest.java
```diff
@@ -0,0 +1,293 @@
...
+    _produceService = new ProduceService(createProduceServiceConfig(props), name);
+    _consumeService = new ConsumeService(createConsumeServiceConfig(props), name);
+    _topicManagementServices = new ArrayList<>();
+    createTopicManagementServices(props, name);
+  }
+
+  private void createTopicManagementServices(Map<String, Object> props, String name) {
+    Map<String, Object> topicManagementConfig = createTopicManagementServiceConfig(props);
+
+    List<String> zookeeperUrlsList = getList(topicManagementConfig, ConsumeServiceConfig.ZOOKEEPER_CONNECT_CONFIG);
```
lindong28: `zookeeper.connect` is used by existing service to represent a zookeepr url. It is confusing to use it for a list of zookeeper url.

tests/src/main/java/com/linkedin/kmf/tests/BasicPipelineTest.java
WarrenGreen: After discussing this with offline with @lindong28 and @smccauliff we've decided to keep this approach but to include a config for this app and to move the `consume.props` and `produce.props` there.

tests/src/main/java/com/linkedin/kmf/tests/BasicPipelineTest.java
WarrenGreen: That's a good point. Now that we intend to create a separate config for this app, I can define a new prop name for a list of zookeeper urls.

tests/src/main/java/com/linkedin/kmf/tests/BasicPipelineTest.java
WarrenGreen: Hmm, we'd be using a hashset to detect the duplicates either way so it'd come down to what is clearest to the user. I could see it being more robust to config errors if we throw an exception. I think I will do this.

tests/src/main/java/com/linkedin/kmf/tests/BasicPipelineTest.java
WarrenGreen: Yeah, perhaps appending an index for each service?

lindong28: I think you can use zookeeper url to identify each service in the log.

WarrenGreen: Probably `name:zookeeperUrl`, right?

lindong28: @WarrenGreen Yeah, that sounds good.

lindong28: @WarrenGreen I will review it again after it is updated.

src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -153,10 +153,10 @@ private void consume() throws Exception {
         _sensors._consumeError.record();
         continue;
       }
-      int partition = record.partition();
-      long index = (Long) avroRecord.get(DefaultTopicSchema.INDEX_FIELD.name());
+      int partition = (int) avroRecord.get(DefaultTopicSchema.PARTITION_FIELD.name());
```
WarrenGreen: @lindong28 @smccauliff We need to key the messages regardless to overcome the lack of partition identity. Rather than change the DefaultTopicSchema, could we simply store the expected index by message key? record.key()

lindong28: I am not sure if I understand the reason for "store the expected index by message key". Can you explain it?

lindong28: After thinking more about the problem yesterday, I think there may be problem with identifying partition using message payload. But before I explain my concern, it is probably better if you can think through this as well since you are the owner of this feature. Can you try to answer the following questions:

- How can we use MirroredPipelineMonitor to detect problem in the source cluster?
- How can we use MirroredPipelineMonitor to detect problem in the second cluster?
- Is there any impact if we identify partition using message payload instead using the RecordMetadata? What it change guarantee of any existing metrics?
- What is the problem we are solving? Is there better solution to address the problem?

smccauliff: So @WarrenGreen  is concerned with the source message ending up in the wrong destination partition after mirroring since MM will do round robin when the source message does not have a key. Yes?  Without a key the consumer on the destination topic might see out-of-order messages with respect to the original (topic, partition) in which they were produced.  Is this the problem?

WarrenGreen: So purely using the message payload to identify the partition will still cause the messages to come out-of-order. In the solution submitted here, the source cluster partition is passed through the message payload and also used as the message key. By doing this I am able to enforce order. After submitting this patch, I realized that passing the source cluster partition in the message payload as well as using it as the key is superfluous. We can simply use the source cluster partition as the key and then store expected indexes using the RecordMetadata key. While the currently proposed solution works fine, it can be more efficient and cleaner by simply using the RecordMetadata.

WarrenGreen:
The newest update uses the RecordMetadata key field to propagate the source cluster partition across mirror maker while maintaining ordering.

src/main/java/com/linkedin/kmf/common/Utils.java
```diff
@@ -100,6 +100,21 @@ public static int createMonitoringTopicIfNotExists(String zkUrl, String topic, i
   }
 
   /**
+   * @param zkUrl zookeeper connection url
```
lindong28: Can you remove this method since it is not used anywhere? Maybe you can simply revert the change on this and other files that doesn't have 'effective' change so that the pull request can be cleaner.

src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -154,9 +154,9 @@ private void consume() throws Exception {
         continue;
       }
       int partition = record.partition();
-      long index = (Long) avroRecord.get(DefaultTopicSchema.INDEX_FIELD.name());
+      long index = (long) avroRecord.get(DefaultTopicSchema.INDEX_FIELD.name());
```
lindong28: Have you tested that you can convert Ojbect to long? I am not sure if it works. See http://stackoverflow.com/questions/2351189/cannot-convert-from-type-object-to-long.

src/main/java/com/linkedin/kmf/services/configs/TopicManagementServiceConfig.java
```diff
@@ -76,7 +77,7 @@
               ConfigDef.Importance.LOW, REBALANCE_INTERVAL_MS_DOC)
       .define(PARTITIONS_TO_BROKER_RATIO_CONFIG,
               ConfigDef.Type.DOUBLE,
-              2.0,
+              PARTITIONS_TO_BROKER_RATIO_DEFAULT,
```
lindong28: We typically don't use extra variable for default value. It is better to keep the code consistent.

WarrenGreen: The two options here are to keep code consistent and have this value stored in two separate places or to break the trend and have a more robust program. I think the latter is a better option

src/main/java/com/linkedin/kmf/common/Utils.java
WarrenGreen: It's used MirrorPipelineMonitor.java:89

src/main/java/com/linkedin/kmf/services/ConsumeService.java
WarrenGreen: Ah, yeah. Good catch.

src/main/java/com/linkedin/kmf/apps/MirrorPipelineMonitor.java
```diff
@@ -0,0 +1,448 @@
...
+  public void awaitShutdown() {
+    for (TopicManagementService topicManagementService : _topicManagementServices) {
+      topicManagementService.awaitShutdown();
+    }
+    _produceService.awaitShutdown();
+    _consumeService.awaitShutdown();
+  }
+
+  @SuppressWarnings("unchecked")
+  private Map<String, Object> getProps(Map<String, Object> props, String key) {
```
lindong28: This method seems redundant. It can be replaced by one line code:

```
props.get(key)
```

src/main/java/com/linkedin/kmf/services/configs/TopicManagementServiceConfig.java
smccauliff: At some point we should convert all these into constants.

src/main/java/com/linkedin/kmf/apps/MirrorPipelineMonitor.java
```diff
@@ -0,0 +1,448 @@
...
+    _name = name;
+    _produceService = new ProduceService(createProduceServiceConfig(props), name);
+    _consumeService = new ConsumeService(createConsumeServiceConfig(props), name);
+    _topicManagementServices = new ArrayList<>();
+    createTopicManagementServices(props, name);
+  }
+
+  private void createTopicManagementServices(Map<String, Object> props, String name) {
+    Map<String, Object> topicManagementConfig = createTopicManagementServiceConfig(props);
+    Double partitionsToBrokerRatio;
```
smccauliff: This can be primitive double.

src/main/java/com/linkedin/kmf/apps/configs/MirrorPipelineMonitorConfig.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package com.linkedin.kmf.apps.configs;
+
+public class MirrorPipelineMonitorConfig {
+
+  public static final String CONSUME_SERVICE_CONFIG = "consume.props";
```
lindong28: Can you provide the specification of all these configs, i.e. required from user or not, default value if not required from user, as we did in TopicManagementServiceConfig? The class probably needs to extend `AbstractConfig` so that it can provide methods to read values. Then the methods such as `getDouble` which you defined in the `MirrorPipelineMonitor` won't be necessary.

BTW, in case it is helpful, in the future you may want to take a look at the existing code that does what you intend to do. For example, we already have method that reads partitionToBrokerRatio from user-provided config. And that is done without defining our own getDouble. Then you can decide what is the best choice for your code.

src/main/java/com/linkedin/kmf/apps/MirrorPipelineMonitor.java
```diff
@@ -0,0 +1,448 @@
...
+    if (topicManagementConfig.containsKey(TopicManagementServiceConfig.PARTITIONS_TO_BROKER_RATIO_CONFIG)) {
```
lindong28: You don't need to read `TopicManagementServiceConfig.PARTITIONS_TO_BROKER_RATIO_DEFAULT` directly. `TopicManagementServiceConfig` handled these for you. Actually, that is the purpose of having `TopicManagementServiceConfig`.

You can probably learn how to use `TopicManagementServiceConfig` by looking at `TopicManagementService`. I think the code below may be simpler:

```
Map<String, Object> topicMangementProps = (Map<String, Object>) props.getOrDefault(MirrorPipelineMonitorConfig.TOPIC_MANAGEMENT_SERVICE_CONFIG, new HashMap<String, Object>());
topicMangementProps.put(CommonServiceConfig.TOPIC_CONFIG, props.get(CommonServiceConfig.TOPIC_CONFIG));
TopicManagementServiceConfig config = new TopicManagementServiceConfig(topicMangementProps);
double partitionsToBrokerRatio = config.getDouble(TopicManagementServiceConfig.PARTITIONS_TO_BROKER_RATIO_THRESHOLD);
``

File: src/main/java/com/linkedin/kmf/apps/MirrorPipelineMonitor.java [hunk 13]
```diff
@@ -0,0 +1,448 @@
...
+      partitionsToBrokerRatio = getDouble(topicManagementConfig, TopicManagementServiceConfig.PARTITIONS_TO_BROKER_RATIO_CONFIG);
+    } else {
+      partitionsToBrokerRatio = TopicManagementServiceConfig.PARTITIONS_TO_BROKER_RATIO_DEFAULT;
+    }
+
+    List<String> zookeeperUrlsList = getList(topicManagementConfig, MirrorPipelineMonitorConfig.ZOOKEEPER_CONNECT_LIST_CONFIG);
```
lindong28: Should zookeeperUrlList be a config of MirrorPipelineMonitorConfig instead of TopicManagementServiceConfig?

src/main/java/com/linkedin/kmf/services/configs/TopicManagementServiceConfig.java
smccauliff: I see how this constant is being used now.  So I think the correct usage pattern of the various XConfig objects is to instantiate them and then let them do all the error checking and population of default values.  That means the constant does not need to be public which means it's safe to remove it.  Probably we still need to make all the defaults constants, but that is a patch for different time.

src/main/java/com/linkedin/kmf/apps/MirrorPipelineMonitor.java
```diff
@@ -0,0 +1,448 @@
...
+  private Map<String, Object> getProps(Map<String, Object> props, String key) {
+    if (props.containsKey(key)) {
+      return (Map<String, Object>) props.get(key);
+    }
+
+    return null;
+  }
+
+  @SuppressWarnings("unchecked")
+  private List<String> getList(Map<String, Object> props, String key) {
```
lindong28: You can probably used `getList()` defined in `MirrorPipelineMonitorConfig` after `MirrorPipelineMonitorConfig` extends `AbstractConfig`. See `TopicManagementService` for example usage. Same suggestion for `getDouble()`.

src/main/java/com/linkedin/kmf/common/Utils.java
lindong28: Yes I can see it now. I didn't find this by doing Ctrl-F and search for getBrokerCount() in the web page probably because the MirrorPipelineMonitor didn't load successfully.

src/main/java/com/linkedin/kmf/apps/MirrorPipelineMonitor.java
```diff
@@ -0,0 +1,448 @@
...
+
+      int brokerCount = com.linkedin.kmf.common.Utils.getBrokerCount(zookeeper);
+      int numPartitions = (int) Math.ceil(partitionsToBrokerRatio * brokerCount);
+      maxNumPartitions = Math.max(maxNumPartitions, numPartitions);
+      zookeeperBrokerCount.put(zookeeper, brokerCount);
+    }
+
+    for (String zookeeper: zookeeperUrlsList) {
+      topicManagementConfig.put(ConsumeServiceConfig.ZOOKEEPER_CONNECT_CONFIG, zookeeper);
+      double partitionToBrokerRatio = maxNumPartitions / zookeeperBrokerCount.get(zookeeper);
```
smccauliff: When brokers are added to one cluster or another the ratio might need to change.  A better way to do this might be to supply some kind of additional object to the topic management service that is the policy that describes the desired number of topics.  That would be a different patch.

smccauliff:
Probably this needs use the XConfig classes.  Other changes could be in a different patch.  What do you say @lindong28 ?

src/main/java/com/linkedin/kmf/services/configs/TopicManagementServiceConfig.java
lindong28: @smccauliff I am not if I understand you correctly. Are you suggesting that we should make all defaults a variable? Since this default value only needs to be specified in the config class once, other classes won't need to reference default value explicitly. Then what is the purpose of making assigning default value to a variable?

lindong28:
@smccauliff I think the MirrorPipelineMonitor has a number of redundant methods that can be simplified. Are you suggesting that we can leave it as it is and refactor the code later? What are the "other changes" that can be in a different patch?

src/main/java/com/linkedin/kmf/apps/MirrorPipelineMonitor.java
```diff
@@ -0,0 +1,448 @@
...
+      throw new RuntimeException(_name + ": Must specify a consume service config.");
+    }
+
+    consumeConfig.put(CommonServiceConfig.TOPIC_CONFIG, props.get(CommonServiceConfig.TOPIC_CONFIG));
+
+    return consumeConfig;
+  }
+
+  private Map<String, Object> createTopicManagementServiceConfig(Map<String, Object> props) {
+    Map<String, Object> topicMangementConfig = getProps(props, MirrorPipelineMonitorConfig.TOPIC_MANAGEMENT_SERVICE_CONFIG);
```
WarrenGreen: Wasn't aware of getOrDefault. I like this solution granted warnings are suppressed for unchecked casts.

src/main/java/com/linkedin/kmf/apps/MirrorPipelineMonitor.java
WarrenGreen: Yeah, that makes sense

src/main/java/com/linkedin/kmf/apps/configs/MirrorPipelineMonitorConfig.java
WarrenGreen: This issue with defining these configs is that there is no Map type in ConfigDef

src/main/java/com/linkedin/kmf/apps/MirrorPipelineMonitor.java
WarrenGreen: I'm not sure how much clearer this is to read but the use of config.getDouble is probably worth it. Making the changes now

src/main/java/com/linkedin/kmf/apps/MirrorPipelineMonitor.java
WarrenGreen: Yeah, perhaps it would be better to create TMS with a known number of partitions to maintain rather than a ratio. Then TMS could dynamically calculate this. Created a JIRA ticket for this.

lindong28: I agree. That can be in a different patch.

src/main/java/com/linkedin/kmf/apps/configs/MirrorPipelineMonitorConfig.java
lindong28: Right. Therefore you can need read map object directly instead of using the config class.

src/main/java/com/linkedin/kmf/apps/MirrorPipelineMonitor.java
lindong28: Maybe you can add a TODO here in the code.

src/main/java/com/linkedin/kmf/apps/configs/MirrorPipelineMonitorConfig.java
WarrenGreen: I'm following the same strategy used for PRODUCER_PROPS_CONFIG

WarrenGreen: Agreed. So consume_service_config, produce_service_config, topic_management_service_config will not be defined while zookeeper_connect_list_config will be defined.

src/main/java/com/linkedin/kmf/apps/MirrorPipelineMonitor.java
```diff
@@ -0,0 +1,410 @@
...
+  private final List<String> _zookeeperUrls;
+  private final String _name;
+
+  public MirrorPipelineMonitor(Map<String, Object> props, String name) throws Exception {
+    _name = name;
+    MirrorPipelineMonitorConfig config = new MirrorPipelineMonitorConfig(props);
+    _zookeeperUrls = config.getList(MirrorPipelineMonitorConfig.ZOOKEEPER_CONNECT_LIST_CONFIG);
+    _produceService = new ProduceService(createProduceServiceConfig(props), name);
+    _consumeService = new ConsumeService(createConsumeServiceConfig(props), name);
+    _topicManagementServices = new ArrayList<>();
```
lindong28: How about `_topicManagementServices = createTopicManagementServices(props, name)`?

Also, maybe rename `createProduceServiceConfig` to `createProduceServiceProps` and `createConsumeServiceConfig` to `createConsumeServiceProps`. Same for `createTopicManagementServiceConfig`.

src/main/java/com/linkedin/kmf/apps/MirrorPipelineMonitor.java
```diff
@@ -0,0 +1,410 @@
...
+ * allows users to track pipeline performance, e.g. latency and throughput.
+ */
+
+public class MirrorPipelineMonitor implements App {
+  private static final Logger LOG = LoggerFactory.getLogger(MirrorPipelineMonitor.class);
+
+  private final ProduceService _produceService;
+  private final ConsumeService _consumeService;
+  private final List<TopicManagementService> _topicManagementServices;
+  private final List<String> _zookeeperUrls;
```
lindong28: Instead of passing `_zookeeperUrls` as class private variable, you can just read it from props in `createTopicManagementServices()`.

src/main/java/com/linkedin/kmf/apps/MirrorPipelineMonitor.java
```diff
@@ -0,0 +1,410 @@
...
+      double partitionToBrokerRatio = maxNumPartitions / zookeeperBrokerCount.get(zookeeper);
+      topicManagementProps.put(TopicManagementServiceConfig.PARTITIONS_TO_BROKER_RATIO_CONFIG, partitionToBrokerRatio);
+      _topicManagementServices.add(new TopicManagementService(topicManagementProps, name + ":" + zookeeper));
+    }
+  }
+
+  @SuppressWarnings("unchecked")
+  private Map<String, Object> createProduceServiceConfig(Map<String, Object> props) {
+    Map<String, Object> produceServiceConfig = (Map<String, Object>) props.getOrDefault(MirrorPipelineMonitorConfig.PRODUCE_SERVICE_CONFIG,
+                                                                                        new HashMap<>());
```
lindong28: Minor point: remove empty line?

src/main/java/com/linkedin/kmf/apps/MirrorPipelineMonitor.java
```diff
@@ -0,0 +1,410 @@
...
+    produceServiceConfig.put(CommonServiceConfig.TOPIC_CONFIG, props.get(CommonServiceConfig.TOPIC_CONFIG));
+
+    return produceServiceConfig;
+  }
+
+  @SuppressWarnings("unchecked")
+  private Map<String, Object> createConsumeServiceConfig(Map<String, Object> props) {
+    Map<String, Object> consumeConfig = (Map<String, Object>) props.getOrDefault(MirrorPipelineMonitorConfig.CONSUME_SERVICE_CONFIG,
+                                                                                 new HashMap<>());
+
```
lindong28: Minor point: remove empty line here?

src/main/java/com/linkedin/kmf/apps/MirrorPipelineMonitor.java
```diff
@@ -0,0 +1,410 @@
...
+    consumeConfig.put(CommonServiceConfig.TOPIC_CONFIG, props.get(CommonServiceConfig.TOPIC_CONFIG));
+
+    return consumeConfig;
+  }
+
+  @SuppressWarnings("unchecked")
+  private Map<String, Object> createTopicManagementServiceConfig(Map<String, Object> props) {
+    Map<String, Object> topicManagementConfig = (Map<String, Object>) props.getOrDefault(MirrorPipelineMonitorConfig.TOPIC_MANAGEMENT_SERVICE_CONFIG,
+      new HashMap<String, Object>());
+
```
lindong28: Minor point: remove empty line?

src/main/java/com/linkedin/kmf/apps/configs/MirrorPipelineMonitorConfig.java
```diff
@@ -0,0 +1,48 @@
...
+
+  public static final String ZOOKEEPER_CONNECT_LIST_CONFIG = "zookeeper.connect.list";
+  public static final String ZOOKEEPER_CONNECT_LIST_DOC = "List of zookeeper connection strings";
+
+  public Double getDouble(String key) {
+    return (Double) get(key);
+  }
+
+  static {
+    CONFIG = new ConfigDef().define(ZOOKEEPER_CONNECT_LIST_CONFIG,
```
lindong28: Can you use the same indentation scheme used in other config classes?

src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -63,8 +63,8 @@
 
   public ConsumeService(Map<String, Object> props, String name) throws Exception {
     _name = name;
-    Map consumerPropsOverride = props.containsKey(ConsumeServiceConfig.CONSUMER_PROPS_CONFIG) ?
-        (Map) props.get(ConsumeServiceConfig.CONSUMER_PROPS_CONFIG) : new HashMap<>();
+    Map consumerPropsOverride = props.containsKey(ConsumeServiceConfig.CONSUMER_PROPS_CONFIG)
+      ? (Map) props.get(ConsumeServiceConfig.CONSUMER_PROPS_CONFIG) : new HashMap<>();
```
lindong28: The existing code style seems better to me. The idea is that when we read `?` at the end of first line, it is more obvious that something else will follow. Can we keep the existing style?

src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -271,12 +271,11 @@ public double measure(MetricConfig config, long now) {
             if (new Double(recordsDelayedRate).isNaN())
               recordsDelayedRate = 0;
 
-            double consumeAvailability = recordsConsumedRate + recordsLostRate > 0 ?
-              (recordsConsumedRate - recordsDelayedRate) / (recordsConsumedRate + recordsLostRate) : 0;
+            double consumeAvailability = recordsConsumedRate + recordsLostRate > 0
+              ? (recordsConsumedRate - recordsDelayedRate) / (recordsConsumedRate + recordsLostRate) : 0;
```
lindong28: Same as above.

smccauliff:
@lindong28 Just my suggestion:  the improvement that needs to be made before the PR can be merged is to use the XConfig classes rather than the methods that exist there already since that reuses the configuration code and is consistent with how the code is configured elsewhere.  Most of the other changes are either minor or are too large (e.g. better coordination of number of partitions).  The larger problems should be fixed with some other PR.

src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -271,12 +271,11 @@ public double measure(MetricConfig config, long now) {
...
 
             return consumeAvailability;
-          }
-        }
+          }        }
```
lindong28: Can we keep the existing code style here?

src/main/java/com/linkedin/kmf/services/configs/TopicManagementServiceConfig.java
```diff
@@ -44,9 +44,9 @@
       + "replication factor used.";
 
   public static final String TOPIC_CREATION_ENABLED_CONFIG = "topic-management.topicCreationEnabled";
-  public static final String TOPIC_CREATION_ENABLED_DOC = "When true this automatically creates the topic mentioned by \"" +
-      CommonServiceConfig.TOPIC_CONFIG + "\" with replication factor \"" + TOPIC_REPLICATION_FACTOR_CONFIG + "and min ISR of max(" +
-      TOPIC_REPLICATION_FACTOR_CONFIG + "-1, 1) with number of brokers * \"" + PARTITIONS_TO_BROKER_RATIO_CONFIG + "\" partitions.";
+  public static final String TOPIC_CREATION_ENABLED_DOC = "When true this automatically creates the topic mentioned by \""
```
lindong28: The existing style seems better to me. But having `+` at the end of previous line it is more obvious to reader that something else will follow in the next line. If there is no obvious reason to change code style, we usually keep the existing style to avoid back-and-forth changes.

lindong28:
@WarrenGreen Thanks for all the effort to make the code better!

src/main/java/com/linkedin/kmf/services/ConsumeService.java
WarrenGreen: I was following LinkedIn style guidelines but while it is a LinkedIn project, it is opensource so perhaps it can diverge. What do you think?

src/main/java/com/linkedin/kmf/apps/MirrorPipelineMonitor.java
WarrenGreen: _zookeeperUrls is a part of MirrorPipelineMonitor props so in order to read in createTopicManagementServices() I'd have to re-parse the props into a MirrorPipelineMonitorConfig. Since _zookeeperUrls is a property of MirrorPipelineMonitor it feels natural to have it as a global variable and follows suite to how ProduceService and ConsumeService handle this.

lindong28:
Thanks! LGTM. I just realized that `props.getOrDefault` is only available in Java 8. I will submit pull request to make it Java 7 compatible.

src/main/java/com/linkedin/kmf/apps/MirrorPipelineMonitor.java
lindong28: I don't think you have to make it a global variable because you already pass the entire properties from user to the parameter of `createTopicManagementServices()`. You can take a look at my next pull request.

lindong28: On another thought, I think it is better to pass zookeeperUrl and topic directly to createProduceServiceProps() etc.

PR #43, opened 2016-12-22
Topic factory

smccauliff:
ATopicFactory interface and a default implementation.  This interface defines what is needed when a topic is automatically created.  Users can define different implementations.  

This change documents the KMProducer and KMConsumer interfaces.


smccauliff:
Forgot to rebase before submitting this PR.

PR #44, opened 2016-12-22
User configurable topic factory to create the monitored topic.

smccauliff:
Also some documentation changes to KMBaseConsumer and KMBaseProducer.

### DIFF
--- a/services/src/main/java/com/linkedin/kmf/common/TopicFactory.java
+++ b/services/src/main/java/com/linkedin/kmf/common/TopicFactory.java
@@ -9,11 +9,14 @@
  */
 package com.linkedin.kmf.common;
 
+import java.util.Properties;
+
+
 /**
  * Constructs the monitored topic if it does not exist.
  *
  * Implementations of this class should have a public constructor with the following signature: <br/>
- *   Constructor(Map<String, ?> config) where config are additional configuration parameters passed in from the Kafka
+ *   Constructor(Map&lt;String, ?&gt; config) where config are additional configuration parameters passed in from the Kafka
  *   Monitor configuration.
  */
 public interface TopicFactory {
@@ -24,10 +27,12 @@ public interface TopicFactory {
    * @param topic topic name
    * @param replicationFactor the replication factor for the topic
    * @param partitionToBrokerRatio This is multiplied by the number brokers to compute the number of partitions in the topic.
+   * @param topicProperties topic specific properties, for example where you would set min.insync.replicas.  Unfortunately
+   *                        the underlying createTopic() uses Properties and not Map&lt;String, ?&gt;.
    * @return The number of partitions for the specified topic.
    */
 
-  public int createTopicIfNotExist(String zkUrl, String topic, int replicationFactor,
-    double partitionToBrokerRatio);
+  int createTopicIfNotExist(String zkUrl, String topic, int replicationFactor,
+    double partitionToBrokerRatio, Properties topicProperties);
 
 }
--- a/services/src/main/java/com/linkedin/kmf/common/TopicFactory.java
+++ b/services/src/main/java/com/linkedin/kmf/common/TopicFactory.java
@@ -29,6 +29,6 @@ public interface TopicFactory {
    * @param config additional configuration parameters passed in from the Kafka Monitor configuration.
    * @return The number of partitions for the specified topic.
    */
-  public int createTopic(String zkUrl, String topic, int replicationFactor,
+  public int createTopicIfNotExist(String zkUrl, String topic, int replicationFactor,
     double partitionToBrokerRatio, Map<String, Object> config);
 }
--- a/services/src/main/java/com/linkedin/kmf/common/TopicFactory.java
+++ b/services/src/main/java/com/linkedin/kmf/common/TopicFactory.java
@@ -1,0 +1,34 @@
+/**
+ * Copyright 2016 LinkedIn Corp. Licensed under the Apache License, Version 2.0 (the "License"); you may not use this
+ * file except in compliance with the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.linkedin.kmf.common;
+
+import java.util.Map;
+
+
+/**
+ * Constructs the monitored topic if it does not exist.
+ *
+ * Implementations of this class should have a public, no-arg constructor.
+ *
+ */
+public interface TopicFactory {
+
+  /**
+   * Creates the specified topic if it does not exist.
+   * @param zkUrl zookeeper connection url
+   * @param topic topic name
+   * @param replicationFactor the replication factor for the topic
+   * @param partitionToBrokerRatio This is multiplied by the number brokers to compute the number of partitions in the topic.
+   * @param config additional configuration parameters passed in from the Kafka Monitor configuration.
+   * @return The number of partitions for the specified topic.
+   */
+  public int createTopic(String zkUrl, String topic, int replicationFactor,
+    double partitionToBrokerRatio, Map<String, Object> config);
+}
--- a/services/src/main/java/com/linkedin/kmf/common/TopicFactory.java
+++ b/services/src/main/java/com/linkedin/kmf/common/TopicFactory.java
@@ -29,6 +29,6 @@ public interface TopicFactory {
    * @param config additional configuration parameters passed in from the Kafka Monitor configuration.
    * @return The number of partitions for the specified topic.
    */
-  public int createTopicIfNotExist(String zkUrl, String topic, int replicationFactor,
+  public int createTopic(String zkUrl, String topic, int replicationFactor,
     double partitionToBrokerRatio, Map<String, Object> config);
 }
--- a/services/src/main/java/com/linkedin/kmf/common/TopicFactory.java
+++ b/services/src/main/java/com/linkedin/kmf/common/TopicFactory.java
@@ -9,14 +9,12 @@
  */
 package com.linkedin.kmf.common;
 
-import java.util.Map;
-
-
 /**
  * Constructs the monitored topic if it does not exist.
  *
- * Implementations of this class should have a public, no-arg constructor.
- *
+ * Implementations of this class should have a public constructor with the following signature: <br/>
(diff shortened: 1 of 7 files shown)


services/src/main/java/com/linkedin/kmf/services/configs/ProduceServiceConfig.java
```diff
@@ -122,7 +129,12 @@
                                     2.0,
                                     atLeast(1.0),
                                     ConfigDef.Importance.LOW,
-                                    CommonServiceConfig.PARTITIONS_TO_BROKER_RATIO_DOC);
+                                    CommonServiceConfig.PARTITIONS_TO_BROKER_RATIO_DOC)
+                            .define(TOPIC_FACTORY_CONFIG,
```
lindong28: To be consistent with the config `consme.consumer.class`, can we use ConfigDef.Type.STRING as type and DefaultTopicFactory.class.getCanonicalName() as default value?

One of the reason `consume.consumer.class` is of type STRING is because user can simply specify OldConsumer and NewConsumer as config value instead of specifying the full class path, which seems slightly easier for user. This can be user here since in the future we may have more default factory class in kafka-monitor.

services/src/main/java/com/linkedin/kmf/common/TopicFactory.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+package com.linkedin.kmf.common;
+
+import java.util.Map;
+
+
+/**
+ * Constructs the monitored topic if it does not exist.
+ *
+ * Implementations of this class should have a public, no-arg constructor.
```
lindong28: Should we require this class to have a constructor with properties as well? For example, when we create a security-enabled topic, do we need to provide the topic factory with url of the security service which hosts authentication information?

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
```diff
@@ -104,9 +106,10 @@ public ProduceService(Map<String, Object> props, String name) throws Exception {
 
     if (existingPartitionCount <= 0) {
       if (config.getBoolean(ProduceServiceConfig.TOPIC_CREATION_ENABLED_CONFIG)) {
+        TopicFactory topicFactory = config.getConfiguredInstance(ProduceServiceConfig.TOPIC_FACTORY_CONFIG, TopicFactory.class);
         _partitionNum.set(
-            Utils.createMonitoringTopicIfNotExists(_zkConnect, _topic, topicReplicationFactor,
-                partitionsToBrokersRatio));
+            topicFactory.createTopic(_zkConnect, _topic, topicReplicationFactor,
+                partitionsToBrokersRatio, props));
```
lindong28: I think it is useful to have props in the signature of createTopic(...). I am wondering if we should leave it as null here.

services/src/main/java/com/linkedin/kmf/common/TopicFactory.java
```diff
@@ -0,0 +1,34 @@
...
+  /**
+   * Creates the specified topic if it does not exist.
+   * @param zkUrl zookeeper connection url
+   * @param topic topic name
+   * @param replicationFactor the replication factor for the topic
+   * @param partitionToBrokerRatio This is multiplied by the number brokers to compute the number of partitions in the topic.
+   * @param config additional configuration parameters passed in from the Kafka Monitor configuration.
+   * @return The number of partitions for the specified topic.
+   */
+  public int createTopic(String zkUrl, String topic, int replicationFactor,
```
radai-rosenblatt: wouldnt it be better to throw if it exists? what if the replication factor arg is different than the existing one?, or any other config?

lindong28: Yeah, if we name it createTopic, I think it is better to throw exception if it exits.

When kafka-monitor was first written, the topic will always be created if it doesn't exist. For convenience of caller I made a method called createTopicIfNotExist() instead of first get the partition number of the topic and then create topic if topic doesn't exist. Now that we have a config to specify whether we need to create the topic, it seems cleaner to just throw exception here.

services/src/main/java/com/linkedin/kmf/services/ProduceService.java
smccauliff: On master I thought the topic creation was still in the ProduceService and so the parameters will be there.  Warren has not yet completed the PR to move the topic creation into the TopicManagementService .  When that is done properties should always be non-null and populated.

services/src/main/java/com/linkedin/kmf/services/configs/ProduceServiceConfig.java
```diff
@@ -53,14 +55,19 @@
 
   public static final String TOPIC_REPLICATION_FACTOR_CONFIG = "topic-management.replicationFactor";
   public static final String TOPIC_REPLICATION_FACTOR_DOC = "When a topic is created automatically this is the "
-      + "replication factor used.";
+    + "replication factor used.";
+
+  public static final String TOPIC_FACTORY_CONFIG = "topic-management.topic-factory";
```
smccauliff: I will make those changes.

services/src/main/java/com/linkedin/kmf/services/configs/ProduceServiceConfig.java
smccauliff: Type CLASS seems like the correct type to use since it's actually a class name.  The user of the configuration needs to know how to convert the easy to use constants into the class names and it needs to be documented.  The configuration class we use has a nice method that just gives the caller an instance of the specified object.   So it's one line of code vs: 1 line to check its s special constant an "else" if it's not.  When it is a special constant then this needs to be converted into the class name (or just  instantiate directly); 3 lines of code.  I say users can suffer with specifying a fully qualified class name.

services/src/main/java/com/linkedin/kmf/common/TopicFactory.java
smccauliff: Kafka seems to follow the pattern of having these plugin classes have a no-arg constructor and method like void configure(Map<String, ?>).  I do like just being able to call newInstance() or the config class's equivalent method.  It also decouples the configuration of a class from it's initialization which might be important if "final" methods need to be assigned some value, but the class needs to be configured later.  I don't see a case where that might happen here.  The other approach ( Constructor(Map<String, ?> ) ), has the advantage of constructing valid objects which is something constructors should enforce when possible.   I don't know what do you think?

In order to construct the ACL an internal implementation of TopicFactoryneeds:
1) Topic Name
2) Principals (which can be extracted from the identity file)
3) Cluster Type
4) Fabric type
5) The datavault uri

Some of these properties would already be present in the TopicManagementService and in the produce service or the consume service configuration.

services/src/main/java/com/linkedin/kmf/common/TopicFactory.java
smccauliff: The produce.topic.topicCreationEnabled is to let Kafka monitor create the topic if it does not exist.  Presumably the user can just leave this set to true and should not get an exception the second time Kafka monitor is run.  I would rather just rename this method to createTopicIfNotExist().  Am I misunderstanding something?

The user can still create the monitored topic outside of Kafka monitor and not having the expected replication factor match the existing replication factor seems to be something that can happen.

services/src/main/java/com/linkedin/kmf/common/TopicFactory.java
lindong28: I agree that if it is a class with util class whose methods are self-contained, then we don't need properties in the constructor. Otherwise, we either need properties in the constructor, or we need a separate init() method in the interface. I think the plugin classes in Kafka also follow this pattern, right?

TopicFactory needs to know the datavault uri in order to create topic. If we don't provide it in TopicFactory constructor or some init() method, then we have to provide it in createTopicIfNotExist(). However, I find it a bit weird for this method to require datavault uri, because datavault uri is actually a config of the factory class instead of per-topic config. What do you think?

services/src/main/java/com/linkedin/kmf/common/TopicFactory.java
```diff
@@ -0,0 +1,33 @@
...
+  /**
+   * Creates the specified topic if it does not exist.
+   * @param zkUrl zookeeper connection url
+   * @param topic topic name
+   * @param replicationFactor the replication factor for the topic
+   * @param partitionToBrokerRatio This is multiplied by the number brokers to compute the number of partitions in the topic.
+   * @return The number of partitions for the specified topic.
+   */
+
+  public int createTopicIfNotExist(String zkUrl, String topic, int replicationFactor,
```
lindong28: It is redundant to specify `public` here. Can you remove it?

I actually think it is useful to have a properties parameter for this method? Currently min.isr = 1. When we move to JBOD we need to use min.isr = 2 when creating topic. We probably want to do it now to avoid changing interface in the future. It is not needed now so we can assign null as its value when this method is called.

services/src/main/java/com/linkedin/kmf/common/TopicFactory.java
smccauliff: I moved the properties into the constructor for the class.  After I did that though I saw the AbstractConfig will call configure(Map<String, ?>) on a new instance it creates.  See getConfiguredInstance in https://github.com/jhooda/kafka/blob/master/clients/src/main/java/org/apache/kafka/common/config/AbstractConfig.java

services/src/main/java/com/linkedin/kmf/services/configs/ProduceServiceConfig.java
```diff
@@ -53,14 +55,19 @@
 
   public static final String TOPIC_REPLICATION_FACTOR_CONFIG = "topic-management.replicationFactor";
   public static final String TOPIC_REPLICATION_FACTOR_DOC = "When a topic is created automatically this is the "
-      + "replication factor used.";
+    + "replication factor used.";
+
+  public static final String TOPIC_FACTORY_CONFIG = "topic-management.topicFactory";
```
lindong28: Should we make this topic-management.topicFactory.class.name?

services/src/main/java/com/linkedin/kmf/common/TopicFactory.java
smccauliff: I added Properties to this method and to the Utils method.

services/src/main/java/com/linkedin/kmf/services/configs/ProduceServiceConfig.java
smccauliff: Made this change.

lindong28:
Thanks for the patch. LGTM!

PR #47, opened 2017-01-05
Rename interface Test to App

lindong28:
This patch has made the following changes:
- Rename interface Test to App
- Rename class BasicEndToEndTest to SingleClusterMonitor
- Rename TopicManagementTest to TopicManagementServiceTest
- Move all sub-projects to root
- Move DefaultTopicFactory and TopicFactory into package com.linkedin.kmf.topicfactory
- Change package of class KafkaMonitor from com.linkedin.kmf.core to com.linkedin.kmf
- Add script bin/single-cluster-monitor
- Deprecate interface com.linkedin.kmf.tests.Test
- Deprecate class com.linkedin.kmf.tests.BasicEndToEndTest
- Update version to 1.0.0


lindong28:
@smccauliff I am not sure that we will not have subproject in the future. For example, we can have a JAVA-based GUI for kafka-monitor as sub-project. It will be useful to keep the core as subproject to leave this option open. Since we have moved all code into subproject `core`, we can still use a single jar file from this subproject, right?

smccauliff:
We could plan for more organization of the code but we should only add the additional complexity  as needed.  It's not needed at this time.  The more build logic you have the more miserable you will be.

If the core subproject ends up sticking around despite my objections then the version field in build.gradle file for subprojects needs to be updated.

lindong28:
@smccauliff I see. Sure, I just moved subprojects to root. BTW, I found that the version is specified in build.gradle twice. Are they both needed? I removed the first one and the `gradle build` successfully generated the right jar. Can you review it again?

smccauliff:
LGTM.  This change greatly simplifies the build logic; makes the package and class naming less confusing.  When testing is completed it should be merged with master.

lindong28:
@smccauliff Thanks! I have successfully tested it using 3 scripts in the bin directory. Will merge it with master now.

PR #48, opened 2017-01-13
KafkaMonitor should map class under Test package into App interface

lindong28:
If user specifies a class under test package in the kafka-monitor.properties, KafkaMonitor should instantiate an instance of this class and map it to the App interface. The patch fixed a bug where such class will be instantiated and mapped into the Service interface.


lindong28:
@smccauliff @WarrenGreen Can you review this patch?

lindong28:
@smccauliff I am not sure if it is necessary to make a tag for every commit. Can we do something similar to Kafka, e.g. make release (or tag) every several weeks or when there is big feature, and with summary of change?

PR #49, opened 2017-01-19
Remove Java-8 only API usage to be Java 7 compatible

lindong28:


lindong28:
@smccauliff @WarrenGreen Can you take a look?

smccauliff:
LGTM.  Thanks for fixing this.  Doesn't the automated build catch these problems though?  We shouldn't be merging PRs that don't build.

lindong28:
@smccauliff Thanks for the review. The automated build has caught this problem. But since Warren is in a hurry to commit the patch and I have some other improvements to make (e.g. read topic using config class), I choose to do it myself in this patch.

PR #51, opened 2017-02-03
Pass the configuration parameters to the topic factory.

smccauliff:
Set the initial delay for the topic management factory thread to be zero.
Let the topic factory deal with existing topic.

### DIFF
--- a/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
+++ b/src/main/java/com/linkedin/kmf/services/TopicManagementService.java
@@ -14,10 +14,8 @@ import com.linkedin.kmf.services.configs.CommonServiceConfig;
 import com.linkedin.kmf.services.configs.TopicManagementServiceConfig;
 import com.linkedin.kmf.topicfactory.TopicFactory;
 import java.io.IOException;
-import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -164,13 +162,12 @@ public class TopicManagementService implements Service  {
     public void run() {
       try {
         TopicState topicState = topicState();
-        if (topicState == null) {
-          if (!_topicCreationEnabled) {
-            throw new RuntimeException(_serviceName + ": topic, " + _topic + ", does not exist and topic creation is not enabled.");
-          }
-
-          topicState = createTopic();
+        
+        if (topicState == null && !_topicCreationEnabled) {
+          throw new RuntimeException(_serviceName + ": topic, " + _topic + ", does not exist and topic creation is not enabled.");
         }
+
+        topicState = createTopic();
 
         int currentReplicationFactor = topicState.replicationFactor();
         if (currentReplicationFactor < 0) {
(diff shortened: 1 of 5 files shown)


src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -248,7 +245,8 @@ public TopicManagementService(Map<String, Object> props, String serviceName) {
     _executor = Executors.newSingleThreadScheduledExecutor(new TopicManagementServiceThreadFactory());
     _configuredTopicReplicationFactor = config.getInt(TopicManagementServiceConfig.TOPIC_REPLICATION_FACTOR_CONFIG);
     _topicCreationEnabled = config.getBoolean(TopicManagementServiceConfig.TOPIC_CREATION_ENABLED_CONFIG);
-    _topicFactoryConfig = config.getString(TopicManagementServiceConfig.TOPIC_FACTORY_CONFIG);
+    String topicFactoryConfig = config.getString(TopicManagementServiceConfig.TOPIC_FACTORY_CONFIG);
+    _topicFactory = (TopicFactory) Class.forName(topicFactoryConfig).getConstructor(Map.class).newInstance(props);
```
lindong28: `props` is the properties of `TopicManagementService`. If user were to pass any config for the factory class, it should probably be specified as an additional map inside `props`. How about we add a config under `TopicManagementServiceConfig`, nam `topic-management.topicFactory.props`, and use its value to construct TopicFactory?

src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -164,14 +162,13 @@ int replicationFactor() {
...
-            throw new RuntimeException(_serviceName + ": topic, " + _topic + ", does not exist and topic creation is not enabled.");
-          }
-
-          topicState = createTopic();
+        
+        if (topicState == null && !_topicCreationEnabled) {
+          throw new RuntimeException(_serviceName + ": topic, " + _topic + ", does not exist and topic creation is not enabled.");
         }
 
+        topicState = createTopic();
```
lindong28: createTopic() is not consistent with what it is doing. It seems more readable and concise to do `topicFactory.createTopicIfNotExist` here directly and remove `createTopic()` method.

I am also wondering whether topicState may be null after this point. If yes, then we will see NPE.

How about doing this:

```
if (_topicCreationEnabled)
  topicFactory.createTopicIfNotExist(...)

TopicState topicState = topicState();

if (topicState == null)
  throw new RuntimeException(...) 
```

I realized that we need `setUncaughtExceptionHandler` in all services to catch exception thrown out of `run()`. I can submit a patch to do it later.

smccauliff: NPE can't happen because createTopic() returns a new topicState or else throws an exception if the topic was not created.  The code that you propose is still more clear so I will make that change.

src/main/java/com/linkedin/kmf/services/TopicManagementService.java
smccauliff: Having nested configuration that matches the composition of classes leaks implementation details into the configuration.  Also it's not clear what the implementations of TopicFactory may need in the future.  Limiting this to the properties seen by the topic management service is probably safe.

lindong28:
Thanks for patch. LGTM.

smccauliff:
Thanks Dong.  I will merge it when the travis build has finished.

smccauliff:
Travis seems stalled to me and I don't know how to unstall it.  Merging.

PR #52, opened 2017-03-09
Synchronize KafkaMonitor

smccauliff:
KafkaMonitor is not synchronized which can lead to strange exceptions when the container it sits in shutsdown.


smccauliff:
Issues with merging pulls in stuff that should not be here.

PR #53, opened 2017-03-09
Catch AdminOperationException

smccauliff:
Otherwise the topic management service can exit.

### DIFF
--- a/src/main/java/com/linkedin/kmf/services/ProduceService.java
+++ b/src/main/java/com/linkedin/kmf/services/ProduceService.java
@@ -141,7 +141,7 @@ public class ProduceService implements Service {
     producerProps.putAll(_producerPropsOverride);
 
     _producer = (KMBaseProducer) Class.forName(_producerClassName).getConstructor(Properties.class).newInstance(producerProps);
-    LOG.info("Producer is initialized.");
+    LOG.info(_name + ": Producer is initialized.");
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


src/main/java/com/linkedin/kmf/services/ProduceService.java
```diff
@@ -141,7 +141,7 @@ private void initializeProducer() throws Exception {
     producerProps.putAll(_producerPropsOverride);
 
     _producer = (KMBaseProducer) Class.forName(_producerClassName).getConstructor(Properties.class).newInstance(producerProps);
-    LOG.info("Producer is initialized.");
+    LOG.info(_name + ": Producer is initialized.");
```
lindong28: I noticed that we already have `LOG.info(_name + ": produce service is initialized.")` in the constructor of `ProduceService`. Maybe we can remove this `LOG.INFO` statement and update the `LOG.info("New partitions added to monitoring.")` in `NewPartitionHandler.run()` to include `_name`.

lindong28: Also, `_name` doesn't necessarily identify the service. For example, bot `ProduceService` and `ConsumeService` of `SingleClusterMonitor` will have the same name. We probably need to include the service name in the log message, e.g. `      LOG.info(_name + "/ProduceService started")`. I should have noticed this earlier. Since you are improving the log message, can you update those message in `ProduceService` and `TopicManagementService` use the format similar to `LOG.info(_name + "/ServiceName do something")`?

smccauliff: There might be a better way to configure the loggers to add what we need to them.  I would rather explore that in a different PR.

smccauliff: Also not catching this exception kills the topic management service in some of our clusters so I would like to have this change merged soon.

lindong28: @smccauliff OK. Let's merge it now in the interest of fixing PROD issue.

PR #54, opened 2017-03-09
Synchronize KafkaMonitor methods

smccauliff:
This protects the monitor state when it is run in a container.

### DIFF
--- a/src/main/java/com/linkedin/kmf/KafkaMonitor.java
+++ b/src/main/java/com/linkedin/kmf/KafkaMonitor.java
@@ -13,12 +13,13 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.linkedin.kmf.services.Service;
 import com.linkedin.kmf.apps.App;
 import com.linkedin.kmf.tests.Test;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import java.io.BufferedReader;
 import java.io.FileReader;
-import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.concurrent.Executors;
@@ -34,15 +35,18 @@ public class KafkaMonitor {
   private static final Logger LOG = LoggerFactory.getLogger(KafkaMonitor.class);
   public static final String CLASS_NAME_CONFIG = "class.name";
 
-  private final Map<String, App> _tests;
-  private final Map<String, Service> _services;
+  /** This is concurrent because healthCheck() can modify this map, but awaitShutdown() can be called at any time by
+   * a different thread.
+   */
+  private final ConcurrentMap<String, App> _apps;
+  private final ConcurrentMap<String, Service> _services;
   private final ScheduledExecutorService _executor;
-  /** When true this instance of the monitor is shutdown and can not be restarted. */
+  /** When true start has been called on this instance of Kafka monitor. */
   private final AtomicBoolean _isRunning = new AtomicBoolean(false);
 
   public KafkaMonitor(Map<String, Map> testProps) throws Exception {
-    _tests = new HashMap<>();
-    _services = new HashMap<>();
+    _apps = new ConcurrentHashMap<>();
+    _services = new ConcurrentHashMap<>();
 
     for (Map.Entry<String, Map> entry : testProps.entrySet()) {
       String name = entry.getKey();
@@ -54,7 +58,7 @@ public class KafkaMonitor {
       if (className.startsWith(App.class.getPackage().getName()) ||
           className.startsWith(Test.class.getPackage().getName())) {
         App test = (App) Class.forName(className).getConstructor(Map.class, String.class).newInstance(props, name);
-        _tests.put(name, test);
+        _apps.put(name, test);
       } else {
         Service service = (Service) Class.forName(className).getConstructor(Map.class, String.class).newInstance(props, name);
         _services.put(name, service);
@@ -67,7 +71,7 @@ public class KafkaMonitor {
     if (!_isRunning.compareAndSet(false, true)) {
       return;
     }
-    for (Map.Entry<String, App> entry: _tests.entrySet()) {
+    for (Map.Entry<String, App> entry: _apps.entrySet()) {
       entry.getValue().start();
     }
     for (Map.Entry<String, Service> entry: _services.entrySet()) {
@@ -89,7 +93,7 @@ public class KafkaMonitor {
   }
 
   private void checkHealth() {
-    Iterator<Map.Entry<String, App>> testIt = _tests.entrySet().iterator();
+    Iterator<Map.Entry<String, App>> testIt = _apps.entrySet().iterator();
     while (testIt.hasNext()) {
       Map.Entry<String, App> entry = testIt.next();
       if (!entry.getValue().isRunning()) {
@@ -113,14 +117,14 @@ public class KafkaMonitor {
       return;
     }
     _executor.shutdownNow();
-    for (App test: _tests.values())
+    for (App test: _apps.values())
       test.stop();
     for (Service service: _services.values())
       service.stop();
   }
 
   public void awaitShutdown() {
-    for (App test: _tests.values())
+    for (App test: _apps.values())
       test.awaitShutdown();
     for (Service service: _services.values())
       service.awaitShutdown();
--- a/src/main/java/com/linkedin/kmf/KafkaMonitor.java
+++ b/src/main/java/com/linkedin/kmf/KafkaMonitor.java
@@ -24,6 +24,11 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
+
+/**
+ * This is the main entry point of the monitor.  It reads the configuration and manages the life cycle of the monitoring
+ * applications.
+ */
 public class KafkaMonitor {
   private static final Logger LOG = LoggerFactory.getLogger(KafkaMonitor.class);
   public static final String CLASS_NAME_CONFIG = "class.name";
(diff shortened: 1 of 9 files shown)


src/main/java/com/linkedin/kmf/KafkaMonitor.java
```diff
@@ -24,18 +24,32 @@
...
+  private boolean _isShutdown = false;
 
-  public KafkaMonitor(Map<String, Map> testProps) throws Exception {
+  public KafkaMonitor() {
     _tests = new HashMap<>();
     _services = new HashMap<>();
+    _executor = Executors.newSingleThreadScheduledExecutor();
+  }
 
+  public synchronized void configure(Map<String, Map> testProps) throws Exception {
```
lindong28: Can you explain the reason of having `configure(...)`? Do we expect this method to be called multiple times?

src/main/java/com/linkedin/kmf/KafkaMonitor.java
```diff
@@ -98,19 +114,26 @@ private void checkHealth() {
     }
   }
 
-  public void stop() {
+  public synchronized void stop() {
```
lindong28: I am not sure we need to check `_isShutdown ` in `stop()` and `awaitShutdown()`. Can you explain the reason of this check?

src/main/java/com/linkedin/kmf/KafkaMonitor.java
```diff
@@ -52,11 +66,13 @@ public KafkaMonitor(Map<String, Map> testProps) throws Exception {
         _services.put(name, service);
       }
     }
-
-    _executor = Executors.newSingleThreadScheduledExecutor();
   }
 
-  public void start() {
+  public synchronized void start() {
+    if (_isShutdown) {
```
lindong28: If the motivation of this check is to prevent `start()` from being invoked more than once, should we do something similar to `ProduceService.start()`?

src/main/java/com/linkedin/kmf/KafkaMonitor.java
smccauliff: In our current container there is some case where the container hits a situation where it needs to shutdown early on, possibly during construction of the application.  The constructor can't be synchronized but a method that can be synchronized.  No, I don't expect this to be called multiple times.

src/main/java/com/linkedin/kmf/KafkaMonitor.java
smccauliff: Since you mention this.  Start should only be called once regardless of the first call to start being successful.  Stop and await shutdown should be allowed to be called multiple times, just incase start or shutdown did not complete successfully the first time.  ProduceService and ConsumeService already have something to prevent their start logic from being executed multiple times.

src/main/java/com/linkedin/kmf/KafkaMonitor.java
smccauliff: See the previous comment.

lindong28: To clarify my previous question, would anything go wrong as of now (without this patch) if user calls `stop()` and `awaitShutdown()` multiple times? It seems fine to me because both `Test.stop()` and `Test.awaitShutdown()` (similarly for `Service`) should be idempotent and allow multiple calls.

src/main/java/com/linkedin/kmf/KafkaMonitor.java
lindong28: The current code in the constructor, including what you have put in the `configure(..)` is non-blocking and should be take very little time to execute. Thus I am still not sure why we have to move that code into a synchronized `configure(...)`.. To phrase the question another way, what would go wrong if we don't have this synchronzied `configure(...)`?

src/main/java/com/linkedin/kmf/KafkaMonitor.java
lindong28: If we assume that `start()` can be called only once, and we will only call `stop()` or `awaitShutdown()` after we have called `start()`, then it seems that `_isShutdown` should always be false when it is checked in `start()`, right? In other words we don't need this check.

src/main/java/com/linkedin/kmf/KafkaMonitor.java
smccauliff: Occasionally we get ConcurrentModificationException on the KafkaMonitor data structures.  It's not clear which thread is interacting with the thread getting the exception.  So the only way I see to make this correct is to synchronize all the methods that use the internal data structures.

src/main/java/com/linkedin/kmf/KafkaMonitor.java
smccauliff: We don't know how other people will integrate this with their own container.  It makes sense to me to just make the default behavior safe.

lindong28: @smccauliff Sure, I actually agree that we should take care of the general case. Then in the most general case, let's assume people can call all three methods in any order concurrently. Then the current approach is not thread safe. I suppose we intend to prevent the body of `stop()` from being executed multiple times by checking `_isShutdown` at its beginning. Yet it can still be executed multiple times if multiple threads call `stop()` concurrently. It seems that we don't want to execute the body of `start()` if user has finished the body of `shutdown()` by checking `istShutdown` at the beginning of `start()`, still it is possible that the body of `start()` can be executed multiple times for similar reason.

Thus it seems that we should do something similar to what `ProduceService()` does by using `AtomicBoolean _running`. The code style would also be more consistent. Does this make sense?

src/main/java/com/linkedin/kmf/KafkaMonitor.java
lindong28: This is weird. No other thread should call anything on KafkaMonitor before KafkaMonitor has finished it construction. Can you provide the exception stack trace? It may be better to understand the problem before fixing it.

smccauliff:
@lindong28 Can you see if these latest changes addressed your concerns?

src/main/java/com/linkedin/kmf/KafkaMonitor.java
```diff
@@ -52,11 +60,13 @@ public KafkaMonitor(Map<String, Map> testProps) throws Exception {
         _services.put(name, service);
       }
     }
-
     _executor = Executors.newSingleThreadScheduledExecutor();
   }
 
-  public void start() {
+  public synchronized void start() {
+    if (!_isRunning.compareAndSet(false, true)) {
```
lindong28: It seem unnecessary to still use AtomaticBoolean if these methods are synchronzied. Can you change `_isRunning` to `volatile boolean` everywhere? Thanks.

smccauliff: AtomicBoolean has the nice method compareAndSet which volatile booleans do not have.  The equivalent code with volatile boolean is longer. 
```java
  if (!_isRunning) {
     _isRunning = true
     //stop implementation here
  }
```
So that's a few extra lines of code.

src/main/java/com/linkedin/kmf/KafkaMonitor.java
smccauliff: So awaitShutdown() should be safe to call multiple times and should be thread safe.  awaitShutdown() does not need to check the started state.  stop also should be safe to call multiple times.  The problem is that healthCheck(), stop() and awaitShutdown() can all get called concurrently. It's not clear to me what stop() should do if start() was not called.  Since stop() can be called in a exit handling thread it's possible it can be called before start().  Should we prevent start from running if stop() was called before start()?

lindong28: I think it is OK for stop() to be called before start(). In that case stop can just return and succeed immediately, right? If we have use-case in which this approach doesn't work, we can throw exception.

src/main/java/com/linkedin/kmf/services/Service.java
```diff
@@ -9,15 +9,39 @@
...
+ * monitoring.
+ */
 public interface Service {
 
   static final String JMX_PREFIX = "kmf.services";
 
+  /**
+   * The start logic must only execute once.  If an error occurs then the implementer of this class must assume that
+   * stop() will be called to clean up.  This method must be thread safe and must assume that stop() may be called
+   * concurrently. This can happen if the monitoring application's life cycle is being managed by a container.  Start
```
lindong28: What does container mean in the open source world? Do you mean the multi-product container?

src/main/java/com/linkedin/kmf/KafkaMonitor.java
lindong28: Then can you change `TopicManagementService` to make the code consistent?

src/main/java/com/linkedin/kmf/KafkaMonitor.java
lindong28: Ah I get your previous question wrong. You are asking about start() not stop().

I would recommend to make the implementation reasonable and simple until there is use-case to justify a more complex semantics. Therefore I think start() means, if not running, then initialize state and set _isrunning to true; and stop() means, if runnning, then tear down everything and set _isrunning to false. If user calls stop() before start(), then stop() is no-op and start() will start things. User needs to call stop() after start() to stop the Kafka Monitor. Does this make sense?

src/main/java/com/linkedin/kmf/services/Service.java
smccauliff: Minimally a container is the thing that owns "main".  But usually containers do other things  such as: object life cycle management, logging and configuration.  Examples of containers include: jboss, spring, pico container.

src/main/java/com/linkedin/kmf/KafkaMonitor.java
smccauliff: Ok.  I think the current state of the code meets all your requirements then.

lindong28:
@smccauliff Can you check another comment I left? The `volatile boolean _running` in `TopicManagementService` is inconsistent with the definition of `_isRunning` in other classes. Also, can you rebase the patch and test it locally? Thanks.

smccauliff:
@lindong28 Rebased.  This has been tested locally by starting two brokers and then monitoring it.  Two new unit tests have also been added.

smccauliff:
I refactored TopicManagementService to use the pattern you suggested.

lindong28:
@smccauliff Awesome. Thanks! LGTM.

PR #56, opened 2017-03-17
Add graphite reporter

ccl0326:
I want to use this project with graphite, so i add a simple implement for it.
If it's not match the required, i'm glad to rewrite it.


build.gradle
```diff
@@ -4,7 +4,7 @@ allprojects {
   apply plugin: 'java'
   apply plugin: 'checkstyle'
 
-  sourceCompatibility = 1.7
+  sourceCompatibility = 1.8
```
lindong28: I think probably should still keep sourceCompatibility = 1.7 until Apache Kafka has changed sourceCompatibility to 1.8.

config/kafka-monitor.properties
```diff
@@ -67,7 +67,10 @@
       "kmf.services:type=consume-service,name=*:records-delay-ms-avg",
       "kmf.services:type=produce-service,name=*:produce-error-rate",
       "kmf.services:type=consume-service,name=*:consume-error-rate"
-    ]
+    ],
+    "report.graphite.host": "localhost",
```
lindong28: The `kafka-monitor.properties` is supposed to be used as a quick-start example for user to try kafka-monitor. Ideally we want to keep it simple and include the most commonly used configs. I am not sure most user will use `graphte`. Maybe we can leave it out of the default config here?

services/src/main/java/com/linkedin/kmf/services/DefaultMetricsReporterService.java
```diff
@@ -32,13 +38,19 @@
   private final List<String> _metricNames;
   private final int _reportIntervalSec;
   private final ScheduledExecutorService _executor;
+  private final GraphiteClient _graphiteClient;
+  private final String _metricNamePrefix;
 
-  public DefaultMetricsReporterService(Map<String, Object> props, String name) {
+  public DefaultMetricsReporterService(Map<String, Object> props, String name) throws SocketException, UnknownHostException {
```
lindong28: I think you are changing `DefaultMetricsReporterService` so that it output data to both log4j and graphite. It means that user will have to use graphte even if they only needs to see metrics in the log4j output. Can you write e.g. `GraphteMetricsReporterService` that implements the `Service` interface to export metrics to graphite? Maybe the graphte-related configs can be put in `GraphteMetricsReporterServiceConfig`.

lindong28:
@ccl0326 Thanks for the patch. Can you check if the comment above makes sense? Also, can you rebase the patch to resolve conflicts between your patch and the latest kafka-monitor code?

ccl0326:
@lindong28 the pr has been modified. :)

src/main/java/com/linkedin/kmf/services/configs/GraphiteMetricsReporterServiceConfig.java
```diff
@@ -0,0 +1,68 @@
...
+  public static final String REPORT_GRAPHITE_HOST_DOC = "The host of graphite server which GraphiteMetricsReporterService will report the metrics values.";
+
+  public static final String REPORT_GRAPHITE_PORT = "report.graphite.port";
+  public static final String REPORT_GRAPHITE_PORT_DOC = "The port of graphite server which GraphiteMetricsReporterService will report the metrics values.";
+
+  public static final String REPORT_GRAPHITE_PREFIX = "report.graphite.prefix";
+  public static final String REPORT_GRAPHITE_PREFIX_DOC = "The prefix of graphite metric name that will be generated with metric name to report to graphite server.";
+
+  static {
+    CONFIG = new ConfigDef().define(REPORT_METRICS_CONFIG,
```
lindong28: Can you adjust the indention to make it consistent, similar to the indention in `DefaultMetricsReporterServiceConfig`?

src/main/java/com/linkedin/kmf/jmx/Utils.java
```diff
@@ -0,0 +1,50 @@
...
+        MBeanAttributeInfo[] attributeInfos = mBeanInfo.getAttributes();
+        for (MBeanAttributeInfo attributeInfo: attributeInfos) {
+          if (attributeInfo.getName().equals(attributeExpr) || attributeExpr.length() == 0 || attributeExpr.equals("*")) {
+            double value = (Double) server.getAttribute(mbeanName, attributeInfo.getName());
+            values.add(new MbeanAttributeValue(mbeanName.getCanonicalName(), attributeInfo.getName(), value));
+          }
+        }
+      }
+    } catch (Exception e) {
+      LOG.error("", e);
```
lindong28: Are we sure that the exception will identify the `mbeanExpr` and `attributeExpr`? Can we log something like `LOG.error("fail to retrieve value for ${mbeanExpr}:${attributeExpr}", e)`?

src/main/java/com/linkedin/kmf/jmx/Utils.java
```diff
@@ -0,0 +1,50 @@
...
+import javax.management.MBeanServer;
+import javax.management.ObjectName;
+
+/**
+ * JMX utilities.
+ */
+public class Utils {
+  private static final Logger LOG = LoggerFactory.getLogger(Utils.class);
+
+  public static List<MbeanAttributeValue> getMBeanAttributeValues(String mbeanExpr, String attributeExpr) {
```
lindong28: I am not sure if it is worth an addition Utils file for this method. Can we add this method to `com.linkedin.kmf.common.Utils`?

ccl0326:
@lindong28 all fixed. i will merge commit after your code review.

lindong28:
@ccl0326 Thanks for the update. LGTM overall. It is OK if you don't squash. Github allows us to sqush the commits before merging them. But can you rebase the patch onto the latest KafkaMonitor repo? Thanks.

ccl0326:
@lindong28 This branch has been update from origin master, the conflicts is not exists now. Do I need to rebase it now?

lindong28:
A patch was committed yesterday which also touches DefaultMetricsReporterService.java. Your patch probably doesn't conflict wit that. But it is probably safer to rebase your onto the latest origin master before merge. You can do `git pull --rebase https://github.com/linkedin/kafka-monitor.git master`, verify that the code works, and update your patch via `git push -f`.

ccl0326:
@lindong28 you are right, i finish all and i apply the `Synchronize KafkaMonitor methods` into GraphiteMetricsReporterService.java.

lindong28:
@ccl0326 Thanks for the patch! LGTM.

PR #59, opened 2017-03-26
Allow ProduceService to start working sooner if the topic is created after ProduceService has started

lindong28:
Currently if we start KafkaMonitor without first creating the topic, the KafkaMonitor's metric will be NAN for up to 40 seconds. This is because the topic may be created after ProduceService starts and the ProduceService only detect topic creation or partition expansion every 40 seconds. This patch reduces the expected time to wait by scheduling the detection to run 1 seconds after ProduceService starts.

Also, we should catch and swallow ToipcExistsException in the TopicManagementService because consumer in the ConsumeService may create the topic when consumer subscribes to the topic if topic auto creation is enabled.

### DIFF
--- a/src/main/java/com/linkedin/kmf/common/Utils.java
+++ b/src/main/java/com/linkedin/kmf/common/Utils.java
@@ -87,7 +87,6 @@ public class Utils {
     ZkUtils zkUtils = ZkUtils.apply(zkUrl, ZK_SESSION_TIMEOUT_MS, ZK_CONNECTION_TIMEOUT_MS, JaasUtils.isZkSecurityEnabled());
     try {
       if (AdminUtils.topicExists(zkUtils, topic)) {
-        LOG.debug("Monitoring topic \"" + topic + "\" already exists.");
         return getPartitionNumForTopic(zkUrl, topic);
       }
 
@@ -102,9 +101,9 @@ public class Utils {
 
       try {
         AdminUtils.createTopic(zkUtils, topic, partitionCount, replicationFactor, topicConfig, RackAwareMode.Enforced$.MODULE$);
-      } catch (TopicExistsException tee) {
+      } catch (TopicExistsException e) {
         //There is a race condition with the consumer.
-        LOG.debug("Monitoring topic \"" + topic + "\" already exists (caught exception).");
+        LOG.debug("Monitoring topic \"" + topic + "\" already exists", e);
         return getPartitionNumForTopic(zkUrl, topic);
       }
       LOG.info("Created monitoring topic \"" + topic + "\" with " + partitionCount + " partitions, min ISR of "
--- a/src/main/java/com/linkedin/kmf/common/Utils.java
+++ b/src/main/java/com/linkedin/kmf/common/Utils.java
@@ -87,7 +87,7 @@ public class Utils {
     ZkUtils zkUtils = ZkUtils.apply(zkUrl, ZK_SESSION_TIMEOUT_MS, ZK_CONNECTION_TIMEOUT_MS, JaasUtils.isZkSecurityEnabled());
     try {
       if (AdminUtils.topicExists(zkUtils, topic)) {
-        LOG.info("Monitoring topic \"" + topic + "\" already exists.");
+        LOG.debug("Monitoring topic \"" + topic + "\" already exists.");
         return getPartitionNumForTopic(zkUrl, topic);
       }
 
@@ -104,7 +104,7 @@ public class Utils {
         AdminUtils.createTopic(zkUtils, topic, partitionCount, replicationFactor, topicConfig, RackAwareMode.Enforced$.MODULE$);
       } catch (TopicExistsException tee) {
         //There is a race condition with the consumer.
-        LOG.info("Monitoring topic \"" + topic + "\" already exists (caught exception).");
+        LOG.debug("Monitoring topic \"" + topic + "\" already exists (caught exception).");
         return getPartitionNumForTopic(zkUrl, topic);
       }
       LOG.info("Created monitoring topic \"" + topic + "\" with " + partitionCount + " partitions, min ISR of "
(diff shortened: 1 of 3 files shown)


lindong28:
@smccauliff Do you have time to review this patch?

smccauliff:
Looking at this now.

src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -164,8 +165,12 @@ int replicationFactor() {
     public void run() {
       try {
         if (_topicCreationEnabled) {
-          _topicFactory.createTopicIfNotExist(_zkConnect, _topic, _configuredTopicReplicationFactor,
-              _partitionsToBrokerRatio, new Properties());
+          try {
+            _topicFactory.createTopicIfNotExist(_zkConnect, _topic, _configuredTopicReplicationFactor,
+                _partitionsToBrokerRatio, new Properties());
+          } catch (TopicExistsException e) {
```
smccauliff: Topic factories should catch this exception. A method called createTopicIfNotExist() kind of implies that you should be free to call this method without getting exceptions if the topic already exists.

src/main/java/com/linkedin/kmf/services/ProduceService.java
```diff
@@ -150,7 +150,7 @@ private void initializeProducer() throws Exception {
   public void start() {
     if (_running.compareAndSet(false, true)) {
       initializeStateForPartitions();
-      _handleNewPartitionsExecutor.scheduleWithFixedDelay(new NewPartitionHandler(), 40_000, 40_000, TimeUnit.MILLISECONDS);
+      _handleNewPartitionsExecutor.scheduleWithFixedDelay(new NewPartitionHandler(), 1000, 30000, TimeUnit.MILLISECONDS);
```
smccauliff: This could check if the topic exists and have a smaller initial delay if it does else a longer one if it does not.

src/main/java/com/linkedin/kmf/services/TopicManagementService.java
lindong28: Good point. I saw the exception trace in the log but I missed the fact that it is already caught.

Then I think we can reduce the log level from INFO to DEBUG if the topic has already been created. The idea is that user should not see exception trace in the log if user starts KafkaMonitor without first creating the topic.

src/main/java/com/linkedin/kmf/services/ProduceService.java
lindong28: I think it is OK to just use 1 second delay by default. If we add this check, the only benefit is that if the topic is already created, the first execution of this task can be later. But this saving doesn't seem to be worth the the extra logic (or lines of code). What do you think?

smccauliff: It might be worth waiting longer if there are many topics that need to be created concurrently.  In our internal configurations we start many monitors in the same process. And only one topic can be created a time because the controller does it.  Yes?

src/main/java/com/linkedin/kmf/common/Utils.java
```diff
@@ -77,7 +77,7 @@ public static int createMonitoringTopicIfNotExists(String zkUrl, String topic, i
     ZkUtils zkUtils = ZkUtils.apply(zkUrl, ZK_SESSION_TIMEOUT_MS, ZK_CONNECTION_TIMEOUT_MS, JaasUtils.isZkSecurityEnabled());
     try {
       if (AdminUtils.topicExists(zkUtils, topic)) {
-        LOG.info("Monitoring topic \"" + topic + "\" already exists.");
+        LOG.debug("Monitoring topic \"" + topic + "\" already exists.");
```
smccauliff: OK. If you are changing this to debug what do you think about setting ProduceRunnable.run catch Exception log to debug.  I'm getting a flood of EKG tickets from things that are probably failures on the broker side.

src/main/java/com/linkedin/kmf/services/ProduceService.java
lindong28: No.. Topic creation is different from partition reassignment. Multiple topics can be created concurrently. All we need to do to create a topic is to create znode `/brokers/topics/[topic]`. Since different KafkaMonitor should uses different topic, these znodes don't override each other.

smccauliff: How does the cluster metadata get updated with the new topic?

src/main/java/com/linkedin/kmf/common/Utils.java
lindong28: I see. Let's talk to SRE to understand if they want to see exception in this case. They probably don't need it. We can remove it if they do not need it. Otherwise, we can assign these EKG ticket to them. Besides, can you have some example EKG tickets? I would like to see what exception may be thrown there.

src/main/java/com/linkedin/kmf/services/ProduceService.java
lindong28: KafkaProduce or KafkaConsume may periodically retrieve latest cluster metadata from broker, or retrieve it when some topic/partition is not found. Does it matter here?

smccauliff: By having a longer wait time it's more likely that the topic management service will have time to create the topic.   That means fewer questions I need to answer from the SREs when they start the monitors and see something strange.  If you think this is not going to happen often enough then I will defer to your greater experience in this matter.

smccauliff:
LGTM.

smccauliff:
Let's talk to the SREs about log levels.

src/main/java/com/linkedin/kmf/services/ProduceService.java
lindong28: I am not sure if I fully understand your comment. Are you concerned with reducing delay of the first execution from 40 sec to 1 sec? Or are you concerned with reducing the periodic delay from 40 sec to 30 sec? Can you explain a bit more what strange thing may happen with shorter delay?

lindong28: @smccauliff Do you have further comment on this? Or I will merge this pull request.

PR #60, opened 2017-03-27
Create TopicManagementServiceManager 

WarrenGreen:
TopicManagementServiceManager to update cluster partition counts to maintain partition to broker ratio

### DIFF
--- a/src/main/java/com/linkedin/kmf/apps/MirrorPipelineMonitor.java
+++ b/src/main/java/com/linkedin/kmf/apps/MirrorPipelineMonitor.java
@@ -15,13 +15,12 @@ import com.linkedin.kmf.services.DefaultMetricsReporterService;
 import com.linkedin.kmf.services.JettyService;
 import com.linkedin.kmf.services.JolokiaService;
 import com.linkedin.kmf.services.ProduceService;
-import com.linkedin.kmf.services.TopicManagementService;
+import com.linkedin.kmf.services.TopicManagementServiceManager;
 import com.linkedin.kmf.services.configs.CommonServiceConfig;
 import com.linkedin.kmf.services.configs.ConsumeServiceConfig;
 import com.linkedin.kmf.services.configs.DefaultMetricsReporterServiceConfig;
 import com.linkedin.kmf.services.configs.ProduceServiceConfig;
 import com.linkedin.kmf.services.configs.TopicManagementServiceConfig;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
@@ -52,7 +51,7 @@ public class MirrorPipelineMonitor implements App {
 
   private final ProduceService _produceService;
   private final ConsumeService _consumeService;
-  private final List<TopicManagementService> _topicManagementServices;
+  private final TopicManagementServiceManager _topicManagementServiceManager;
   private final String _name;
 
   public MirrorPipelineMonitor(Map<String, Object> props, String name) throws Exception {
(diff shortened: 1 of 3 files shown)


src/main/java/com/linkedin/kmf/services/TopicManagementServiceManager.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class TopicManagementServiceManager implements Service {
```
lindong28: Can we just extend the TopicManagementService to be able to monitor multiple clusters? It seems a bit weird to have a scheduler that is specifically used to update partitionsToBrokerRatio of other services.

WarrenGreen: Yeah, we can probably do that. Things to consider with that though:

1. It would be a major rewrite of TopicManager
2. It would probably still be good to have a class that manages cluster-level granularity which a higher level service pulls in.

smccauliff: Can we have something that detects changes in all the monitored partitions in a pipeline and then just stops the TopicManagementService and starts a new one when the parameters change?  This way the TopicManagementService can just stay the same.

smccauliff: Also it looks like TopicManagementServiceManager now has the functionality that TopicManagementService had.   Will this break all the existing monitoring apps that use this class?

WarrenGreen: This is the creation of TopicManagementServiceManager so nothing previously relied on it. There aren't any major changes to TopicManagementService so all apps should be fine with this.

WarrenGreen: @smccauliff That's kind of what this is doing. It's using TopicManagementService functionality that already exists. It just monitors cluster changes and updates the partitionToBroker ratio periodically. Then the TopicManagementService does what it normally does with this new ratio.

WarrenGreen: @lindong28 Perhaps a better way to do this is to rename it as MultipleClusterTopicManagementService and to extend the TopicManagementService

src/main/java/com/linkedin/kmf/apps/MirrorPipelineMonitor.java
```diff
@@ -52,7 +51,7 @@
 
   private final ProduceService _produceService;
   private final ConsumeService _consumeService;
-  private final List<TopicManagementService> _topicManagementServices;
+  private final TopicManagementServiceManager _topicManagementServiceManager;
   private final String _name;
 
   public MirrorPipelineMonitor(Map<String, Object> props, String name) throws Exception {
```
lindong28: Should this class be named `MutilClusterMonitor` or `MirroredPipelineMonitor`? I just noticed that `MirrorPipelineMonitor` doesn't describe itself very well.

src/main/java/com/linkedin/kmf/services/TopicManagementServiceManager.java
lindong28: @WarrenGreen I think the `MultipleClusterTopicManagementService` is better than `TopicManagementServiceManager`. It is better to have a new service than adding a new concept `manager`. I am OK with not doing it in the existing `TopicManagementService` if we don't duplicate code.

I am not sure it makes sense to have this class extend TopicManagementService. You can try it if you think the code would look nice.

src/main/java/com/linkedin/kmf/services/TopicManagementServiceManager.java
```diff
@@ -0,0 +1,154 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(TopicManagementServiceManager.class);
+
+  private double _partitionsToBrokerRatio;
+  private String _name;
+  private List<String> _zookeeperUrls;
+  private AtomicBoolean _running;
+  private final ScheduledExecutorService _managementExecutor;
+  private Map<String, TopicManagementService> _services;
+
+  public TopicManagementServiceManager(Map<String, Object> props, List<String> zookeeperUrls, String topic,
```
lindong28: If we want to add a new service that manages topic across multiple clusters, can this service has its own config and takes a properties map as parameter of its constructor, just like every other services? It probably needs its own config class.

src/main/java/com/linkedin/kmf/services/TopicManagementServiceManager.java
```diff
@@ -0,0 +1,154 @@
...
+      topicManagementProps.put(TopicManagementServiceConfig.PARTITIONS_TO_BROKER_RATIO_CONFIG, individualPartitionToBrokerRatio);
+      _services.put(zookeeper, new TopicManagementService(topicManagementProps, name + ":" + zookeeper));
+    }
+
+    LOG.info(_name + ": topic management service manager is initialized.");
+  }
+
+  @SuppressWarnings("unchecked")
+  private Map<String, Object> createTopicManagementServiceProps(Map<String, Object> props, String topic) {
+    Map<String, Object> topicManagementServiceProps = props.containsKey(MirrorPipelineMonitorConfig.TOPIC_MANAGEMENT_SERVICE_CONFIG)
```
lindong28: We want each service to be reusable. App can depend on Service's config but not the vise versa. Thus we don't want a service class to reference a config from MirrorPipelineMonitorConfig.

src/main/java/com/linkedin/kmf/services/TopicManagementServiceManager.java
```diff
@@ -0,0 +1,154 @@
...
+  private class TopicManagementServiceManagerRunnable implements Runnable {
+
+    @Override
+    public void run() {
+      Map<String, Integer> brokerCounts = getBrokerCounts(_zookeeperUrls);
+      int maxNumPartitions = getMaxNumberOfPartitions(brokerCounts);
+
+      for (String zookeeper: _zookeeperUrls) {
+        double individualPartitionToBrokerRatio = maxNumPartitions / brokerCounts.get(zookeeper);
+        _services.get(zookeeper).setPartitionsToBrokerRatio(individualPartitionToBrokerRatio);
```
lindong28: I have a few concern with this approach:

- This doesn't necessarily ensure that the partition count will be maxNumPartitions in all clusters. TopicManagementService only expand partition when PARTITIONS_TO_BROKER_RATIO_THRESHOLD is violated.

- It seems hacky to change the config of another class (i.e. TopicManagementService) on the fly. This makes it harder to reason about TopicManagementService and extend it in the future. We should try to avoid this.

- Currently the default value of REBALANCE_INTERVAL_MS_CONFIG is 10 min. It means that the partition count may be inconsistent for up to 10 minutes since TopicMangementService of each cluster runs at different time. Reducing this interval may alleviate the problem but it is not a good idea to rely on a low interval to reduce race condition. Ideally we want a single thread to change partition count in all clusters altogether.

src/main/java/com/linkedin/kmf/services/TopicManagementServiceManager.java
lindong28: @WarrenGreen I think the MultipleClusterTopicManagementService is better than TopicManagementServiceManager. It is better to have a new service than adding a new concept manager. I am OK with not doing it in the existing TopicManagementService if we don't duplicate code.

I am not sure it makes sense to have this class extend TopicManagementService. You can try it if you think the code would look nice.

lindong28: @WarrenGreen I think I just accidentally deleted your last comment.

src/main/java/com/linkedin/kmf/services/TopicManagementServiceManager.java
WarrenGreen: These are good concerns. Let's meet today to discuss them

PR #61, opened 2017-03-27
Automatically generate the configuration documentation in html.

smccauliff:
Automatically generate the configuration documentation in `html`.


build.gradle
```diff
@@ -54,6 +57,25 @@ allprojects {
     configFile = new File(rootDir, "checkstyle/checkstyle.xml")
   }
 
+  task createConfigDocDir() {
+    outputs.dir configDocDir
+    file(configDocDir).mkdirs()
+  }
+
+  task createConfigDocs( dependsOn : ['createConfigDocDir', 'compileJava'], type : JavaExec) {
```
lindong28: It seems that we don't need createConfigDocDir since the ConfigDocumentationGenerator would create the directory if it doesn't exist. Also, does this task need to depend on `compileJava`? I tested the patch without this dependency and it still works.

build.gradle
```diff
@@ -54,6 +57,25 @@ allprojects {
...
+    classpath sourceSets.main.runtimeClasspath
+    main = 'com.linkedin.kmf.common.ConfigDocumentationGenerator'
+    args = [configDocDir,
+            'com.linkedin.kmf.services.configs.ConsumeServiceConfig',
+            'com.linkedin.kmf.services.configs.DefaultMetricsReporterServiceConfig',
+            'com.linkedin.kmf.services.configs.JettyServiceConfig',
+            'com.linkedin.kmf.services.configs.ProduceServiceConfig',
+            'com.linkedin.kmf.services.configs.TopicManagementServiceConfig',
+            'com.linkedin.kmf.apps.configs.MirrorPipelineMonitorConfig']
```
lindong28: Can you rebase the patch and change MirrorPipelineMonitorConfig to MultiClusterMonitorConfig?

build.gradle
```diff
@@ -54,6 +57,25 @@ allprojects {
...
 
+  task createConfigDocDir() {
+    outputs.dir configDocDir
+    file(configDocDir).mkdirs()
+  }
+
+  task createConfigDocs( dependsOn : ['createConfigDocDir', 'compileJava'], type : JavaExec) {
+    classpath sourceSets.main.runtimeClasspath
+    main = 'com.linkedin.kmf.common.ConfigDocumentationGenerator'
+    args = [configDocDir,
```
lindong28: Do you think it would be better to specify this list in the Java code?

build.gradle
```diff
@@ -1,3 +1,6 @@
+def configDocDir = "${buildDir}/configDocs"
```
lindong28: I am wondering if it may be better to have the doc under ${project.rootDir}/docs instead of ${buildDir}. User typically would ignore the content under build directory. On the other hand, the docs under root directory seems more obvious. Apache Kafka generated docs under ${project.rootDir}/docs/generated.

build.gradle
smccauliff: Done.

build.gradle
smccauliff: Doing a clean build becomes more problematic if files generated by the build are not under the build directory.  As it stands now I can clean the build by just removing one directory.  What does having more than one directory buy us?

build.gradle
smccauliff: The code itself does not really depend the functionality of the input classes so adding them as compile time dependencies does not really make any sense.  If the list is wrong the build will still fail.

build.gradle
lindong28: Make sense. Then let's keep it in build directory.

build.gradle
lindong28: Sure.

src/main/java/com/linkedin/kmf/common/ConfigDocumentationGenerator.java
```diff
@@ -0,0 +1,65 @@
...
+      printHelp();
+      System.exit(1);
+    }
+
+    File outputDir = new File(argv[0]);
+    if (!outputDir.exists()) {
+      outputDir.mkdirs();
+    }
+
+    for (int i = 1; i < argv.length; i++) {
```
lindong28: It seems that what you have done here is similar (if not the same) as the implementation of `org.apache.kafka.common.config.ConfigDef.toHtmlTable()`. Can we simply print the string `ConsumeServiceConfig.CONFIG.toHtmlTable()` to a html file for each config class instead of writing new code here to generate the html file? I tried both your patch and `System.out.println(ConsumeServiceConfig.CONFIG.toHtmlTable())`. They actually generate the same output (except the title).

More specifically, I think we can probably add the follow main methods to each subclass of `AbstractConfig` (e.g. `ConsumeServiceConfig`). Then we can invoke these main methods to generate html files.

```
public static void main(String[] args) {
  System.out.println(CONFIG.toHtmlTable());
}
```

src/main/java/com/linkedin/kmf/services/MultiClusterTopicManagementService.java
```diff
@@ -182,6 +182,7 @@ public void run() {
     private final double _minPartitionsToBrokersRatio;
     private final int _minPartitionNum;
     private final TopicFactory _topicFactory;
+    private final Properties _topicProperties;
```
lindong28: Thanks for updating the commit message of the other pull request. Because that forced push has changed the hash of that commit, can you rebase this pull request onto the latest master so that we won't include the changes in `MultiClusterTopicManagementService.java` and `TopicManagementServiceConfig.java` as part of this pull request?

andrewchoi5:
Is this active?

smccauliff:
I ran out  of time to work on this.   Are you interested in  reviewing it?

andrewchoi5:
Completed resolving the `merge` conflicts.

src/main/java/com/linkedin/kmf/common/ConfigDocumentationGenerator.java
andrewchoi5: We can add this feature/revisit in a follow-up PR.

PR #63, opened 2017-04-07
Remove dash from schema

lindong28:


lindong28:
Remove dash from scheme name so that it can be parsed by avro 1.7.

smccauliff:
I can't think of any problems this would cause unless someone had an existing consumer that was far behind in consuming messages, but that's actually a real problem if that is happening.  Did you test this?  When you merge squash and merge can you change the description so it says avro 1.7 rather than 1.4.

lindong28:
This should not cause issue because consumer used by Kafka Monitor will always consume from latest offset. I have tested it and it works. Why should we change the description to say avro 1.7? It seems that avro 1.4 already has this requirement. It is just not enforced until avro 1.7.

smccauliff:
But we are making this change for 1.7 compatibility.    Someone reading this may want to know that.

lindong28:
@smccauliff Personally I don't think it matters. Because we should make this change anyway since it is specified by avro 1.4 and later. But it doesn't really matters. I have changed the description to 1.7.

smccauliff:
Remove dash from scheme name so that it meets the naming conversion of avro library.

The schema name should meet the following requirement according to avro 1.7 specification (https://avro.apache.org/docs/1.7.7/spec.html#Names)

start with [A-Za-z_]
subsequently contain only [A-Za-z0-9_]

smccauliff:
Thanks for dealing with the avro upgrade complexities.

lindong28:
@smccauliff Thanks for the review.

PR #64, opened 2017-04-11
MultiClusterTopicManagementService implementation and various improvements

lindong28:
New feature:
- Implemented MultiClusterTopicManagementService to make sure that monitor topics across clusters have the same number of partitions
- Implemented MultiClusterMonitor to monitor a pipeline of Kafka clusters

Improve logging and simplify operation:
- KafkaMonitor will log error message every 5 seconds if any service or app is not running
- MultiClusterTopicManagementService will increase replication factor of the monitor topic if it is smaller than the configured replication
- ConsumeService will wait for 1 sec before running to reduce the chance that it creates topic before TopicManagementService

Simplify config:
- Removed the config topic-management.partitionsToBrokersThreshold

Code cleanup and reafactor:
- Fixed typo in kafka-monitor.properties to use partitionsToBrokersRatio instead of partitionsToBrokerRatio
- Improved log to avoid String concatenation
- Removed duplicated code in EndToEndTest and marked it as deprecated
- Renamed MirrorPipelineMonitor to MultiClusterMonitor
- Refactored TopicManagementService to replace TopicState class with private static methods


src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -118,7 +118,7 @@ public void run() {
           LOG.error(_name + "/ConsumeService failed", e);
         }
       }
-    }, _name + " consume-service");
+    }, _name + " consume-service-thread");
```
smccauliff: I think I had the string "thread" in the thread names at some point and we decided that it can make the thread names too long.

src/main/java/com/linkedin/kmf/services/JolokiaService.java
```diff
@@ -36,14 +36,14 @@ public JolokiaService(Map<String, Object> props, String name) throws Exception {
   public synchronized void start() {
     if (_isRunning.compareAndSet(false, true)) {
       _jolokiaServer.start();
-      LOG.info(_name + "/JolokiaService started at port 8778");
+      LOG.info("{}/JolokiaService started at port 8778", _name);
```
smccauliff: The port is always 8778?

src/main/java/com/linkedin/kmf/services/MultiClusterTopicManagementService.java
```diff
@@ -0,0 +1,343 @@
...
+          }
+        }
+      } catch (Exception e) {
+        LOG.error(_serviceName + "/MultiClusterTopicManagementService will stop due to error.", e);
+        stop();
+      }
+    }
+  }
+
+  private static void maybeCreateTopic(Map props) throws Exception {
```
smccauliff: It would be better if these methods (the methods that manipulate the topic state in Kafka) were not static so it would be easier to overload and test them.

src/main/java/com/linkedin/kmf/services/MultiClusterTopicManagementService.java
```diff
@@ -0,0 +1,343 @@
...
+    String zkConnect = config.getString(CommonServiceConfig.ZOOKEEPER_CONNECT_CONFIG);
+    Boolean topicCreationEnabled = config.getBoolean(TopicManagementServiceConfig.TOPIC_CREATION_ENABLED_CONFIG);
+    int replicationFactor = config.getInt(TopicManagementServiceConfig.TOPIC_REPLICATION_FACTOR_CONFIG);
+    double partitionsToBrokerRatio = config.getDouble(TopicManagementServiceConfig.PARTITIONS_TO_BROKERS_RATIO_CONFIG);
+    String topicFactoryClassName = config.getString(TopicManagementServiceConfig.TOPIC_FACTORY_CLASS_CONFIG);
+    Map topicFactoryConfig = props.containsKey(TopicManagementServiceConfig.TOPIC_FACTORY_PROPS_CONFIG) ?
+        (Map) props.get(TopicManagementServiceConfig.TOPIC_FACTORY_PROPS_CONFIG) : new HashMap();
+    TopicFactory topicFactory = (TopicFactory) Class.forName(topicFactoryClassName).getConstructor(Map.class).newInstance(topicFactoryConfig);
+
+    if (topicCreationEnabled) {
```
smccauliff: If this check happens towards the beginning of the method then the user is spared from needing to have valid values for the other configuration items.

src/main/java/com/linkedin/kmf/services/JolokiaService.java
lindong28: Yes it is hardcoded as 8778.. It is intended for demo but not PROD deployment. So it is probably OK.

src/main/java/com/linkedin/kmf/services/ConsumeService.java
lindong28: I don't remember this discussion.. I will remove this from the thread name.

src/main/java/com/linkedin/kmf/services/MultiClusterTopicManagementService.java
lindong28: We can certainly make the methods non-static. But I think the static methods make code easier to interpret because it is very clear which input parameters will generate output. How about we make these methods non-static later when we need to overload them?

src/main/java/com/linkedin/kmf/services/MultiClusterTopicManagementService.java
lindong28: I think the current code is more concise than if we read the value of topic-management.topicCreationEnabled, do this check, and read values of other configs, right? I am not sure the benefit of allowing users to have invalid configs if topic creation is disabled. If I am user I would want to be notified the first time I specified an invalid config value. What do you think?

src/main/java/com/linkedin/kmf/services/JolokiaService.java
smccauliff: OK

src/main/java/com/linkedin/kmf/services/MultiClusterTopicManagementService.java
smccauliff: These methods also have Map as their single parameter.  These methods are not configuration methods so why do they have only Map as a parameter?  The way it is coded now it's not clear what any of the input parameters are supposed to be.

src/main/java/com/linkedin/kmf/services/MultiClusterTopicManagementService.java
smccauliff: Usually if something is disabled then the dependent checks are also disabled.  For example if SSL/TLS is not being used by the Kafka clients then the Kafka clients don't check that the user has specified a key store and other properties it uses for secure communication.

src/main/java/com/linkedin/kmf/services/MultiClusterTopicManagementService.java
lindong28: This props is the same props that we pass to the constructor of TopicManagementService. But I can not pass TopicManagementServiceConfig because we need to get configs such as TOPIC_FACTORY_PROPS_CONFIG directly from the Map. Alternative we can pass all parameters (e.g. topic, zkconnect) as parameters to these methods but it makes the signature of these methods very clumsy. What would you recommend?

src/main/java/com/linkedin/kmf/services/MultiClusterTopicManagementService.java
lindong28: @smccauliff It is fixed now. Can you take another look?

src/main/java/com/linkedin/kmf/services/MultiClusterTopicManagementService.java
```diff
@@ -146,198 +158,198 @@ public void run() {
...
-    int replicationFactor = partitionInfoList.get(0).replicas().length;
-    for (PartitionInfo partitionInfo : partitionInfoList) {
-      if (replicationFactor != partitionInfo.replicas().length) {
-        String topic = partitionInfoList.get(0).topic();
-        throw new RuntimeException("Partitions of the topic " + topic + " have different replication factor");
+      int replicationFactor = partitionInfoList.get(0).replicas().length;
+      for (PartitionInfo partitionInfo : partitionInfoList) {
+        if (replicationFactor != partitionInfo.replicas().length) {
+          String topic = partitionInfoList.get(0).topic();
+          throw new RuntimeException("Partitions of the topic " + topic + " have different replication factor");
```
smccauliff: This would cause the topic management service to exit.  Do we want that that?

lindong28: Yes, I think it is OK to exit. It is actually impossible to see different replication factor for partitions of the same topic if the topic is created using our current API and tool. It is not an intermediate state that can recovery by itself later. Thus TopicManagementService needs to stop for someone to look into the problem.

src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -10,417 +10,57 @@
...
-    while (it.hasNext()) {
-      scala.Tuple2<Object, scala.collection.Seq<Object>> scalaTuple = it.next();
-      Integer partition = (Integer) scalaTuple._1();
-      scala.Option<Object> leaderOption = _zkUtils.getLeaderForPartition(_topic, partition);
-      Node leader = leaderOption.isEmpty() ?  null : new Node((Integer) leaderOption.get(), "", -1);
-      Node[] replicas = new Node[scalaTuple._2().size()];
-      for (int i = 0; i < replicas.length; i++) {
-        Integer brokerId = (Integer) scalaTuple._2().apply(i);
-        replicas[i] = new Node(brokerId, "", -1);
+  private Map<String, Object> createMultiClusterTopicManagementServiceProps(Map<String, Object> props, String serviceName) {
```
smccauliff: It's hard to know if this method is correct. Having some documentation, especially about the output, would be helpful.

lindong28: Sure. How about the following documentation:

  /**
   * @param props a map of key/value pair used for configuring TopicManagementService
   * @param serviceName service name
   * @return a map of the following format:
   *
   * {
   *   "topic.management.props.per.cluster" : {
   *     // all key/value pair from props except one with key "topic"
   *   }
   *   "topic": topic
   * }
   */

lindong28: @smccauliff The comment is added. Thanks for your review!

src/main/java/com/linkedin/kmf/services/MultiClusterTopicManagementService.java
smccauliff: OK

smccauliff:
Once the CI checks are complete you can squash and merge this.  Thanks for completing this refactor.

PR #67, opened 2017-05-03
Update kafka-monitor-start.bat to use the correct class path

lindong28:


lindong28:
@smccauliff Can you review this? Thanks.

smccauliff:
run-class.bat emits some kind of error message that references the old class name

smccauliff:
Did you actually test this on windows?

lindong28:
@smccauliff You are right. More change is needed in kmf-run-class.bat. I have updated the patch to fix it. I haven't tested it on Windows since I don't have window machine. I think it is probably better than not fixing it. We can ask that user to test it after the fix. What do you think?

PR #68, opened 2017-05-04
Allow topic-management.partitionsToBrokersRatio to be smaller than 1

lindong28:


lindong28:
There is use-case where user may want to set partitionsToBrokerRatio to be smaller than 1. For example, when we monitor kafka-rest endpoint, we want the partition number to be 1 regardless of the number of brokers because kafka-rest may not allow user to produce with key.

lindong28:
@smccauliff Can you please review this?

PR #70, opened 2017-05-11
Added Statsd reporter to export metrics to Graphite.

arsenaid:


lindong28:
@arsenaid Thanks for your patch! Previously we had merged the pull request https://github.com/linkedin/kafka-monitor/pull/56 which added `GraphiteMetricsReporterService` to report metrics to Graphite. Does that patch address the motivation of this patch?

arsenaid:
Hi @lindong28. There are a few reasons I added Statsd export:

1) The existing Graphite exporting doesn't work or I cannot configure it out of box.
2) Statsd is a lightweight metrics aggregator using UDP instead of plain-text or other binary protocols

lindong28:
@arsenaid I see. Ideally we would like to have only one class to be used for sending metrics data to Graphite. Otherwise the code base will get harder to maintain over time. Maybe we should first try to understand why that `GraphiteMetricsReporterService` doesn't work for you and fix that. And I think it is reasonable to replace the existing graphte solution with your solution if the Statsd protocol is strictly better than the protocol used by the other solution. 

BTW, the `statsd-service` is probably not suitable to be put in `kafka-monitor.properties` because `kafka-monitor.properties` is intended to be work out-of-the-box and provides only the basic config. Not all user has Graphite service in their cluster.

The other solution used net.savantly:graphite-client:1.1.0-RELEASE and your solution used com.timgroup:java-statsd-client:3.0.1 to send metrics to graphte. Is there a standard or dominant library used to send metrics to graphite?

@ccl0326 Do you have time to review this patch and comment on the choice of the library? Any comment on the comparison between Statsd and the protocol used in your patch? Thanks.

arsenaid:
StatsD can integrate with different backends. Graphite's carbon is the most common integration used in many companies. 

Here is a link to the Stack Overflow thread on the reasoning for using StatsD aggregation. http://stackoverflow.com/questions/13697553/why-use-statsd-when-graphites-carbon-aggregator-can-do-the-same-job

Maybe we can create a separate the Statsd config from the main `kafka-monitor.properties` as an example. That way default config is not affected. What would you suggest?

I don't have any objection on using another StatsD library. I just used the library which is preferred choice in the Spring Boot project. It's a proven one and works out of box with little configuration.

I think supporting different methods for exporting to metrics would be good for the users of the monitor.

lindong28:
@arsenaid Sure, this makes sense. The patch looks good to me. Instead of creating another Statsd config specifically to show how to use and config `StatsdMetricsReporterService` in the properties file, would it work to just provide the example config for `StatsdMetricsReporterService` in `StatsdMetricsReporterServiceConfig`? I am not sure we should have one example config file for each newly added reporter service.

arsenaid:
Sounds good. I moved the config example to the `StatsdMetricsReporterServiceConfig` file.

lindong28:
@arsenaid Thanks for the patch!

ccl0326:
Sorry for late reply.
From the integrate way, statsd is better than graphite. I add graphite reporter is because the metrics can send to storage directly, if use statsd, the user should add middleware like brubeck.

P.S. I think com.timgroup:java-statsd-client is not maintenance now, we use com.indeed.java-dogstatsd-client in production.

PR #73, opened 2017-05-26
Add config topic-management.minPartitionNum

lindong28:


src/main/java/com/linkedin/kmf/services/MultiClusterTopicManagementService.java
```diff
@@ -182,13 +199,13 @@ public void run() {
 
     void maybeCreateTopic() throws Exception {
       if (_topicCreationEnabled) {
-        _topicFactory.createTopicIfNotExist(_zkConnect, _topic, _replicationFactor, _partitionsToBrokerRatio, new Properties());
+        _topicFactory.createTopicIfNotExist(_zkConnect, _topic, _replicationFactor, _minPartitionsToBrokersRatio, new Properties());
```
smccauliff: Did you want to make the change to propagate min isr config?

smccauliff:
LGMT!

src/main/java/com/linkedin/kmf/services/MultiClusterTopicManagementService.java
lindong28: @smccauliff Make sense. This is definitely worth doing. I will submit a separate patch either today or tomorrow to address this problem.

lindong28: @smccauliff Thanks for the fast review!

PR #74, opened 2017-05-27
Support custom implementation of App and Service interfaces

lindong28:


lindong28:
As of now we consider a class to be App if its name in kafka-monitor.properties starts with `com.linkedin.kmf.apps.App`. However, user may want to run a class that implements the App interface under a different package name. This patch addresses this problem by using a better way to identify whether the class implements the App interface.

becketqin:
LGTM.

PR #75, opened 2017-05-28
Add topic-management.topic.props config to specify topic properties

lindong28:


lindong28:
Ping @smccauliff for review.

PR #76, opened 2017-05-29
Add metric to detect offline runnables

lindong28:


lindong28:
Ping @smccauliff for review.

src/main/java/com/linkedin/kmf/KafkaMonitor.java
```diff
@@ -98,15 +123,17 @@ private void checkHealth() {
     while (testIt.hasNext()) {
```
smccauliff: At some point the code here was removing dead tests here using the iterator, but since it's not doing that any more this could just use some kind of for(...) loop.  No?

src/main/java/com/linkedin/kmf/KafkaMonitor.java
```diff
@@ -66,6 +78,19 @@ public KafkaMonitor(Map<String, Map> testProps) throws Exception {
       }
     }
     _executor = Executors.newSingleThreadScheduledExecutor();
+
+    _offlineRunnables = new ConcurrentHashMap<>();
+    List<MetricsReporter> reporters = new ArrayList<>();
+    reporters.add(new JmxReporter(JMX_PREFIX));
```
smccauliff: Did we not have some way of reporting JMX metrics before?

src/main/java/com/linkedin/kmf/KafkaMonitor.java
```diff
@@ -33,12 +42,15 @@
 public class KafkaMonitor {
   private static final Logger LOG = LoggerFactory.getLogger(KafkaMonitor.class);
   public static final String CLASS_NAME_CONFIG = "class.name";
+  private static final String METRIC_GROUP_NAME = "kafka-monitor";
+  private static final String JMX_PREFIX = "kmf";
```
smccauliff: I thought we retired the "kmf" prefix in favor of "kafka-monitor"?

lindong28: We changed the project name from kafka-monitor-framework but the package name is still `com.linkedin.kmf` and the JMX prefix used by all existing metrics are still `kmf.services`. I don't think we will likely change the existing package name and JMX prefix. Thus it may be OK to use `kmf` in the future package name and the JMX prefix so that they are consistent. I am not strong about the JMX prefix name here. Do you prefer me to change it to `kafka-monitor`?

src/main/java/com/linkedin/kmf/KafkaMonitor.java
lindong28: Yes, the existing services (e.g. `ProduceService`) will report JMX metrics using similar code. But `KafkaMonitor.java` doesn't report metrics as of now. Does this answer your question?

src/main/java/com/linkedin/kmf/KafkaMonitor.java
lindong28: I agree. Will update the code to use for loop. Thanks.

src/main/java/com/linkedin/kmf/KafkaMonitor.java
smccauliff: That's fine then I don't want to break an existing code.  Does JMX_PREFIX belong in this file if other metrics also use this prefix?

lindong28: All subclass of `com.linkedin.kmf.services.Service` uses `com.linkedin.kmf.services.JMX_PREFIX` as the JMX prefix, which is currently `kmf.services`. Since the metric we are adding here belong to `com.linkedin.kmf.KafkaMonitor`, I think it may be reasonable to use `kmf` as its JMX prefix.

So the short answer to your question here is that other existing metrics don't use the JMX prefix that we added in this patch.

smccauliff:
LGMT!  Thanks!

PR #77, opened 2017-06-02
Skip partition assignment if partition number < broker number

lindong28:


lindong28:
There is no need to reassign partitions to cover every broker if number of partition is smaller than the number of brokers. This may happen if user topic-management.partitionsToBrokersRatio is smaller than 1.

smccauliff:
LGTM

PR #79, opened 2017-06-23
Docker build hardening

SlevinBE:
- removes some non-existing folders from being added to the Docker image
- properly handle signals in Docker start script so that the Kafka Monitor process is getting stopped when receiving a SIGINT or SIGTERM signal.

### DIFF
--- a/docker/kafka-monitor-docker-entry.sh
+++ b/docker/kafka-monitor-docker-entry.sh
@@ -15,7 +15,13 @@
 
 set -x
 
+# SIGTERM-handler
+trap 'pkill java; exit 130' SIGINT
+trap 'pkill java; exit 143' SIGTERM
+
 #  wait for DNS services to be available
 sleep 10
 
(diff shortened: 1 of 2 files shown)


docker/kafka-monitor-docker-entry.sh
```diff
@@ -15,7 +15,13 @@
 
 set -x
 
+# SIGTERM-handler
+trap 'pkill java; exit 129' SIGINT
```
lindong28: According to https://en.wikipedia.org/wiki/Unix_signal, the port number for SIGINT is 2. Should this be 130 since 128 + 2 = 130?

lindong28:
@SlevinBE oops.. Sorry, I forgot to review this patch. Just left one minor comment.

SlevinBE:
@lindong28 Thanks for reviewing this. I've resolved the comment.

lindong28:
Thanks for the patch @SlevinBE. LGTM.

PR #83, opened 2017-06-29
Keep properties of the monitor topic updated

lindong28:


lindong28:
@smccauliff Can you review this patch?

This patch makes sure that if user specified the topic config are the monitor topic has been created, or if the replication factor is increased after the topic has been created, kafka-monitor will update the topic properties (e.g. min.insync.replicas) if needed.

smccauliff:
LGTM!

PR #85, opened 2017-07-10
Override min.insync.isr for the monitor topic only if it is specified by the user

lindong28:


smccauliff:
This has checkstyle violtions.

lindong28:
@smccauliff My bad. It is fixed now. Thanks!

src/main/java/com/linkedin/kmf/services/MultiClusterTopicManagementService.java
```diff
@@ -259,11 +254,18 @@ void maybeReassignPartitionAndElectLeader() throws Exception {
...
+        // Update the properties of the monitor topic if any config is different from the user-specified config
         Properties currentProperties = AdminUtils.fetchEntityConfig(zkUtils, ConfigType.Topic(), _topic);
-        if (!currentProperties.equals(_topicProperties)) {
+        Properties expectedProperties = new Properties();
+        for (Object key: currentProperties.keySet())
+          expectedProperties.put(key, currentProperties.get(key));
+        for (Object key: _topicProperties.keySet())
+          expectedProperties.put(key, _topicProperties.get(key));
+
+        if (!currentProperties.equals(expectedProperties)) {
```
smccauliff: What happens when the user has not specified some property?

lindong28: Kafka broker is configured with `min.insync.replicas`. If we don't explicitly specify optional properties such as `min.insync.replicas`, then the value in the broker config will be applied. The purpose of this patch is to allow kafka-monitor to use the default config if the Kafka cluster if possible.

lindong28: @smccauliff Thanks for the review! Do you think we can merge this patch?

smccauliff: Won't this trigger everytime a user has a missing property?  The two properties will never be the same.  No?

lindong28: No, it won't trigger every time. Note that I am comparing `currentProperties` with `expectedProperties`, where the `expectedProperties` is the union of the user-specified properties (which may be empty) and the `currentProperties`.

PR #93, opened 2017-08-07
Upgrade kafka dependency to 0.11.0.1

lindong28:


lindong28:
@smccauliff can you take a look at this patch?

smccauliff:
Did you want to increment the version number to 2.0?

smccauliff:
These lines in build.gradle:

  version = "1.0.0"

  sourceCompatibility = 1.7

Kafka 0.11 is also Java 8 compatible so this can be changed to 1.8

src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -224,9 +223,7 @@ public boolean isRunning() {
     private final Sensor _recordsDelay;
     private final Sensor _recordsDelayed;
 
-    public ConsumeMetrics(Metrics metrics, final Map<String, String> tags) {
-      this.metrics = metrics;
-
+    public ConsumeMetrics(final Metrics metrics, final Map<String, String> tags) {
```
smccauliff: Why add final here and not everywhere else it could be added?  IMO final for parameters degrades readability.  If reassigning to parameters is a code quality problem then perhaps a better check style file would be the way to go.

src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -265,9 +262,9 @@ public ConsumeMetrics(Metrics metrics, final Map<String, String> tags) {
         new Measurable() {
           @Override
           public double measure(MetricConfig config, long now) {
-            double recordsConsumedRate = _sensors.metrics.metrics().get(new MetricName("records-consumed-rate", METRIC_GROUP_NAME, tags)).value();
-            double recordsLostRate = _sensors.metrics.metrics().get(new MetricName("records-lost-rate", METRIC_GROUP_NAME, tags)).value();
-            double recordsDelayedRate = _sensors.metrics.metrics().get(new MetricName("records-delayed-rate", METRIC_GROUP_NAME, tags)).value();
+            double recordsConsumedRate = metrics.metrics().get(metrics.metricName("records-consumed-rate", METRIC_GROUP_NAME, tags)).value();
```
smccauliff: I don't know very much about this metrics API so I assume this is the correct thing to do.

src/main/java/com/linkedin/kmf/services/ConsumeService.java
lindong28: It seems that the latest Kafka still use Java 1.7 for source compatibility. And if compilation with Java 1.7 will fail if we do not add `final` here. `final` is only needed here because it is used to measure `consume-availability-avg` in this method.

src/main/java/com/linkedin/kmf/services/ConsumeService.java
lindong28: Yeah it is needed for Kafka 0.11.

PR #95, opened 2017-08-11
ProduceAvailability should be 1.0 as long as ProduceService does not see exception

lindong28:


lindong28:
@smccauliff can you take a look?

PR #109, opened 2018-01-18
Add KafkaMetricsReporterService to support send metrics to kafka

hackerwin7:
A demand to send metrics to Kafka,then other applications could consume from Kafka for further processing or compute.

### DIFF
--- a/config/kafka-monitor.properties
+++ b/config/kafka-monitor.properties
@@ -77,23 +77,23 @@
   },
 
 
-#  "reporter-service": {
-#    "class.name": "com.linkedin.kmf.services.DefaultMetricsReporterService",
-#    "report.interval.sec": 1,
-#    "report.metrics.list": [
-#      "kmf:type=kafka-monitor:offline-runnable-count",
-#      "kmf.services:type=produce-service,name=*:produce-availability-avg",
-#      "kmf.services:type=consume-service,name=*:consume-availability-avg",
-#      "kmf.services:type=produce-service,name=*:records-produced-total",
-#      "kmf.services:type=consume-service,name=*:records-consumed-total",
-#      "kmf.services:type=consume-service,name=*:records-lost-total",
-#      "kmf.services:type=consume-service,name=*:records-duplicated-total",
-#      "kmf.services:type=consume-service,name=*:records-delay-ms-avg",
-#      "kmf.services:type=produce-service,name=*:records-produced-rate",
-#      "kmf.services:type=produce-service,name=*:produce-error-rate",
-#      "kmf.services:type=consume-service,name=*:consume-error-rate"
-#    ]
-#  }
+  "reporter-service": {
+    "class.name": "com.linkedin.kmf.services.DefaultMetricsReporterService",
+    "report.interval.sec": 1,
+    "report.metrics.list": [
+      "kmf:type=kafka-monitor:offline-runnable-count",
+      "kmf.services:type=produce-service,name=*:produce-availability-avg",
+      "kmf.services:type=consume-service,name=*:consume-availability-avg",
+      "kmf.services:type=produce-service,name=*:records-produced-total",
+      "kmf.services:type=consume-service,name=*:records-consumed-total",
+      "kmf.services:type=consume-service,name=*:records-lost-total",
+      "kmf.services:type=consume-service,name=*:records-duplicated-total",
+      "kmf.services:type=consume-service,name=*:records-delay-ms-avg",
+      "kmf.services:type=produce-service,name=*:records-produced-rate",
+      "kmf.services:type=produce-service,name=*:produce-error-rate",
+      "kmf.services:type=consume-service,name=*:consume-error-rate"
+    ]
+  }
 
 
 #  Example statsd-service to report metrics
@@ -118,7 +118,7 @@
 #    "zookeeper.connect": "localhost:2181",
 #    "bootstrap.servers": "localhost:9092",
 #    "topic": "kafka-monitor-topic-metrics",
-#    "report.topic.replication.factor": 1,
+#    "reporter.kafka.topic.replication.factor": 1,
 #    "report.metrics.list": [
 #      "kmf.services:type=produce-service,name=*:produce-availability-avg",
 #      "kmf.services:type=consume-service,name=*:consume-availability-avg",
--- a/config/kafka-monitor.properties
+++ b/config/kafka-monitor.properties
@@ -69,20 +69,20 @@
 
   "reporter-service": {
     "class.name": "com.linkedin.kmf.services.DefaultMetricsReporterService",
-    "report.interval.sec": 1,
-    "report.metrics.list": [
-      "kmf:type=kafka-monitor:offline-runnable-count",
-      "kmf.services:type=produce-service,name=*:produce-availability-avg",
-      "kmf.services:type=consume-service,name=*:consume-availability-avg",
-      "kmf.services:type=produce-service,name=*:records-produced-total",
-      "kmf.services:type=consume-service,name=*:records-consumed-total",
-      "kmf.services:type=consume-service,name=*:records-lost-total",
-      "kmf.services:type=consume-service,name=*:records-duplicated-total",
-      "kmf.services:type=consume-service,name=*:records-delay-ms-avg",
-      "kmf.services:type=produce-service,name=*:records-produced-rate",
-      "kmf.services:type=produce-service,name=*:produce-error-rate",
-      "kmf.services:type=consume-service,name=*:consume-error-rate"
-    ]
+    "report.interval.sec": 1
+#    "report.metrics.list": [
+#      "kmf:type=kafka-monitor:offline-runnable-count",
+#      "kmf.services:type=produce-service,name=*:produce-availability-avg",
+#      "kmf.services:type=consume-service,name=*:consume-availability-avg",
+#      "kmf.services:type=produce-service,name=*:records-produced-total",
+#      "kmf.services:type=consume-service,name=*:records-consumed-total",
+#      "kmf.services:type=consume-service,name=*:records-lost-total",
+#      "kmf.services:type=consume-service,name=*:records-duplicated-total",
+#      "kmf.services:type=consume-service,name=*:records-delay-ms-avg",
+#      "kmf.services:type=produce-service,name=*:records-produced-rate",
+#      "kmf.services:type=produce-service,name=*:produce-error-rate",
+#      "kmf.services:type=consume-service,name=*:consume-error-rate"
+#    ]
   },
 
   "jetty-service": {
--- a/config/kafka-monitor.properties
+++ b/config/kafka-monitor.properties
@@ -118,7 +118,7 @@
 #    "zookeeper.connect": "localhost:2181",
 #    "bootstrap.servers": "localhost:9092",
 #    "topic": "kafka-monitor-topic-metrics",
-#    "reporter.kafka.topic.replication.factor": 1,
+#    "report.kafka.topic.replication.factor": 1,
 #    "report.metrics.list": [
 #      "kmf.services:type=produce-service,name=*:produce-availability-avg",
 #      "kmf.services:type=consume-service,name=*:consume-availability-avg",
--- a/config/kafka-monitor.properties
+++ b/config/kafka-monitor.properties
@@ -117,9 +117,21 @@
 #    "report.interval.sec": 3,
 #    "zookeeper.connect": "localhost:2181",
 #    "bootstrap.servers": "localhost:9092",
-#    "topic": "kafka-monitor-topic-metrics"
+#    "topic": "kafka-monitor-topic-metrics",
+#    "report.metrics.list": [
+#    "kmf:type=kafka-monitor:offline-runnable-count",
+#    "kmf.services:type=produce-service,name=*:produce-availability-avg",
+#    "kmf.services:type=consume-service,name=*:consume-availability-avg",
+#    "kmf.services:type=produce-service,name=*:records-produced-total",
+#    "kmf.services:type=consume-service,name=*:records-consumed-total",
+#    "kmf.services:type=consume-service,name=*:records-lost-total",
+#    "kmf.services:type=consume-service,name=*:records-duplicated-total",
+#    "kmf.services:type=consume-service,name=*:records-delay-ms-avg",
+#    "kmf.services:type=produce-service,name=*:records-produced-rate",
+#    "kmf.services:type=produce-service,name=*:produce-error-rate",
+#    "kmf.services:type=consume-service,name=*:consume-error-rate"
+#    ]
 #  }
-
 
 }
 
--- a/config/kafka-monitor.properties
+++ b/config/kafka-monitor.properties
@@ -44,7 +44,7 @@
   "single-cluster-monitor": {
     "class.name": "com.linkedin.kmf.apps.SingleClusterMonitor",
     "topic": "kafka-monitor-topic",
-    "zookeeper.connect": "localhost:2181/kafka",
+    "zookeeper.connect": "localhost:2181",
     "bootstrap.servers": "localhost:9092",
     "produce.record.delay.ms": 100,
     "topic-management.topicCreationEnabled": true,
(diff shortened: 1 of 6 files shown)


hackerwin7:
fixed some format issues with checkstyle

lindong28:
@hackerwin7 Thanks for the patch! I will review it

src/main/java/com/linkedin/kmf/common/Utils.java
```diff
@@ -108,6 +108,33 @@ public static int createMonitoringTopicIfNotExists(String zkUrl, String topic, i
...
+  /**
+   * Create general topic (for Kafka metrics reporter).
+   *
+   * @param zkUrl zookeeper connection url
+   * @param topic topic name
+   * @param replicationFactor replication factor
+   * @param partitionCount topic partition number
+   * @param topicConfig additional topic configuration
+   */
+  public static void createTopicIfNotExists(String zkUrl, String topic, int replicationFactor,
```
lindong28: Can we reduce the code duplication by reusing the existing `createTopicIfNotExists`? We can either just set partitionToBrokerRatio to be very small fraction number, e.g. 0.00001, or add a new parameter such as minPartitionNum. Then the topic can be created with `createTopicIfNotExists(partitionToBrokerRatio=0, minPartitionNum=1)`. What do you think?

src/main/java/com/linkedin/kmf/services/KafkaMetricsReporterService.java
```diff
@@ -0,0 +1,144 @@
...
+import java.util.Map;
+import java.util.Properties;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Created by IntelliJ IDEA.
+ * User: hackerwin7
+ * Date: 2018/01/18
```
lindong28: Similar to Apache Kafka, we generally do not include this. Can you remove it?

config/kafka-monitor.properties
```diff
@@ -111,6 +111,16 @@
 #  }
 
 
+#  Example kafka-service to report metrics 
+#  "reporter-kafka-service": {
+#    "class.name": "com.linkedin.kmf.services.KafkaMetricsReporterService",
+#    "report.interval.sec": 3,
+#    "zookeeper.connect": "localhost:2181",
```
lindong28: Can we add `report.metrics.list` in this example?

src/main/java/com/linkedin/kmf/services/configs/KafkaMetricsReporterServiceConfig.java
```diff
@@ -0,0 +1,81 @@
...
+    CONFIG = new ConfigDef().define(REPORT_METRICS_CONFIG,
+      ConfigDef.Type.LIST,
+      Arrays.asList("kmf.services:*:*"),
+      ConfigDef.Importance.MEDIUM,
+      REPORT_METRICS_DOC)
+      .define(REPORT_INTERVAL_SEC_CONFIG,
+        ConfigDef.Type.INT,
+        1,
+        ConfigDef.Importance.LOW,
+        REPORT_INTERVAL_SEC_DOC)
```
lindong28: Could you change the indentation to be similar to other config classes?

src/main/java/com/linkedin/kmf/services/configs/KafkaMetricsReporterServiceConfig.java
```diff
@@ -0,0 +1,81 @@
...
+
+import org.apache.kafka.common.config.AbstractConfig;
+import org.apache.kafka.common.config.ConfigDef;
+
+import java.util.Arrays;
+import java.util.Map;
+
+/**
+ * Created by IntelliJ IDEA.
+ * User: hackerwin7
```
lindong28: Could you remove this?

src/main/java/com/linkedin/kmf/services/configs/KafkaMetricsReporterServiceConfig.java
```diff
@@ -0,0 +1,81 @@
...
+  public static final String ZOOKEEPER_CONNECT_CONFIG = CommonServiceConfig.ZOOKEEPER_CONNECT_CONFIG;
+  public static final String ZOOKEEPER_CONNECT_DOC = CommonServiceConfig.ZOOKEEPER_CONNECT_DOC;
+
+  public static final String BOOTSTRAP_SERVERS_CONFIG = CommonServiceConfig.BOOTSTRAP_SERVERS_CONFIG;
+  public static final String BOOTSTRAP_SERVERS_DOC = CommonServiceConfig.BOOTSTRAP_SERVERS_DOC;
+
+  public static final String TOPIC_CONFIG = CommonServiceConfig.TOPIC_CONFIG;
+  public static final String TOPIC_DOC = CommonServiceConfig.TOPIC_DOC;
+
+  public static final String PRODUCER_ID_CONFIG = "produce.producer.id";
```
lindong28: It seems that Apache Kafka producer does not have this config. Could you tell me where is this config used?

src/main/java/com/linkedin/kmf/services/KafkaMetricsReporterService.java
```diff
@@ -0,0 +1,144 @@
...
+    _reportIntervalSec = config.getInt(KafkaMetricsReporterServiceConfig.REPORT_INTERVAL_SEC_CONFIG);
+    _executor = Executors.newSingleThreadScheduledExecutor();
+
+    _brokerList = config.getString(KafkaMetricsReporterServiceConfig.BOOTSTRAP_SERVERS_CONFIG);
+    _producerId = config.getString(KafkaMetricsReporterServiceConfig.PRODUCER_ID_CONFIG);
+    initializeProducer();
+
+    _zkConnect = config.getString(KafkaMetricsReporterServiceConfig.ZOOKEEPER_CONNECT_CONFIG);
+    _topic = config.getString(KafkaMetricsReporterServiceConfig.TOPIC_CONFIG);
+    Utils.createTopicIfNotExists(_zkConnect, _topic, 1, 1, new Properties()); // fixed partition and factor
```
lindong28: If broker number is more than 1, we probably want to set replication factor to be more than 1, right?

src/main/java/com/linkedin/kmf/common/Utils.java
hackerwin7: Thanks your review! Can we set the `createTopicIfNotExists` as primitive method which can be invoke by `createMonitoringTopicIfNotExists`  with `partitionToBrokerRatio` parameter?
Such as:
```
createMonitoringTopicIfNotExists(topic, partitionToBrokerRatio, ...) {
  createTopicIfNotExists(topic, partitionCount, factor, ...)
}
```
Maybe we just need a more general method `createTopicIfNotExists` rather than `createMonitoringTopicIfNotExists` ?

src/main/java/com/linkedin/kmf/services/KafkaMetricsReporterService.java
hackerwin7: Maybe I should put out the `replicationFactor` as a configuration parameter, and `partitionCount` is 
fixed `1` ? either, put out both as configuration ?

config/kafka-monitor.properties
hackerwin7: Okay, I'll modify the sample like this:
```
#  Example kafka-service to report metrics 
#  "reporter-kafka-service": {
#    "class.name": "com.linkedin.kmf.services.KafkaMetricsReporterService",
#    "report.interval.sec": 3,
#    "zookeeper.connect": "localhost:2181",
#    "bootstrap.servers": "localhost:9092",
#    "topic": "kafka-monitor-topic-metrics",
#    "report.metrics.list": [
#    "kmf:type=kafka-monitor:offline-runnable-count",
#    "kmf.services:type=produce-service,name=*:produce-availability-avg",
#    "kmf.services:type=consume-service,name=*:consume-availability-avg",
#    "kmf.services:type=produce-service,name=*:records-produced-total",
#    "kmf.services:type=consume-service,name=*:records-consumed-total",
#    "kmf.services:type=consume-service,name=*:records-lost-total",
#    "kmf.services:type=consume-service,name=*:records-duplicated-total",
#    "kmf.services:type=consume-service,name=*:records-delay-ms-avg",
#    "kmf.services:type=produce-service,name=*:records-produced-rate",
#    "kmf.services:type=produce-service,name=*:produce-error-rate",
#    "kmf.services:type=consume-service,name=*:consume-error-rate"
#    ]
#  }
```

src/main/java/com/linkedin/kmf/services/KafkaMetricsReporterService.java
hackerwin7: Thanks, I'll remove these

src/main/java/com/linkedin/kmf/services/configs/KafkaMetricsReporterServiceConfig.java
hackerwin7: remove these

src/main/java/com/linkedin/kmf/services/configs/KafkaMetricsReporterServiceConfig.java
hackerwin7: Seemed confused to me.  
`TOPIC_CONFIG` is for `KafkaMetricsReporterService` producer send topic.
`PRODUCER_ID_CONFIG` is for `KafkaMetricsReporterService` producer client id when constructed Apache Kafka producer.

src/main/java/com/linkedin/kmf/services/configs/KafkaMetricsReporterServiceConfig.java
hackerwin7: Okay, I'll tune the format.

src/main/java/com/linkedin/kmf/services/KafkaMetricsReporterService.java
lindong28: It think it is reasonable to have replicationFactor config for KafkaMetricsReporterService. I suppose the throughput of the metrics messages should be low. So it is probably OK to fix the partition number to 1.

src/main/java/com/linkedin/kmf/services/configs/KafkaMetricsReporterServiceConfig.java
lindong28: If the purpose of `produce.producer.id` is to be used as client.id of the producer, maybe you can just use the client.id directly? That can makes the config easier to understand.

BTW, is there any reason to explicitly specify client.id for producer here?

src/main/java/com/linkedin/kmf/common/Utils.java
lindong28: Translating `partitionNum` to `partitionToBrokerRatio` requires your code to figure out the broker number and the code will be a little more complex. Yes, I think we can have a more general method createTopicIfNotExists(...) which takes parameters minPartitionToBrokerRatio and minPartitionNum. The partition number of the topic can be determined as max(minPartitionToBrokerRatio * brokerNum, minPartitionNum). Does this sound OK?

hackerwin7: Thanks your reply! So, I'll modify the method `createMonitoringTopicIfNotExists(zkUrl, topic, replicationFactor,partitionToBrokerRatio, topicConfig)` to the method `createTopicIfNotExists(zkUrl, topic, replicationFactor,partitionToBrokerRatio, minPartitionNum, topicConfig)`. is it reasonable for this?

src/main/java/com/linkedin/kmf/services/configs/KafkaMetricsReporterServiceConfig.java
hackerwin7: Okay, maybe a fixed client.id suit for this.

hackerwin7:
Fixed some review issues with the code.

config/kafka-monitor.properties
```diff
@@ -77,23 +77,23 @@
...
-      "kmf.services:type=consume-service,name=*:records-consumed-total",
-      "kmf.services:type=consume-service,name=*:records-lost-total",
-      "kmf.services:type=consume-service,name=*:records-duplicated-total",
-      "kmf.services:type=consume-service,name=*:records-delay-ms-avg",
-      "kmf.services:type=produce-service,name=*:records-produced-rate",
-      "kmf.services:type=produce-service,name=*:produce-error-rate",
-      "kmf.services:type=consume-service,name=*:consume-error-rate"
-    ]
-  }
+#  "reporter-service": {
```
lindong28: Can we include the existing "reporter-service" in the default kafka-monitor.properties? I think most users probably would want to see the metrics in the console by default.

src/main/java/com/linkedin/kmf/services/configs/KafkaMetricsReporterServiceConfig.java
```diff
@@ -0,0 +1,77 @@
...
+  public static final String ZOOKEEPER_CONNECT_CONFIG = CommonServiceConfig.ZOOKEEPER_CONNECT_CONFIG;
+  public static final String ZOOKEEPER_CONNECT_DOC = CommonServiceConfig.ZOOKEEPER_CONNECT_DOC;
+
+  public static final String BOOTSTRAP_SERVERS_CONFIG = CommonServiceConfig.BOOTSTRAP_SERVERS_CONFIG;
+  public static final String BOOTSTRAP_SERVERS_DOC = CommonServiceConfig.BOOTSTRAP_SERVERS_DOC;
+
+  public static final String TOPIC_CONFIG = CommonServiceConfig.TOPIC_CONFIG;
+  public static final String TOPIC_DOC = CommonServiceConfig.TOPIC_DOC;
+
+  public static final String TOPIC_REPLICATION_FACTOR = "report.topic.replication.factor";
```
lindong28: Can we name this config to be reporter.kafka.topic.replication.factor? I am hoping that the first part of the config key could identify the serivce, e.g. KafkaMetricsReporterServiceConfig.

src/main/java/com/linkedin/kmf/services/KafkaMetricsReporterService.java
```diff
@@ -0,0 +1,139 @@
...
+      List<MbeanAttributeValue> attributeValues = getMBeanAttributeValues(mbeanExpr, attributeExpr);
+      for (MbeanAttributeValue attributeValue : attributeValues) {
+        String metric = attributeValue.toString();
+        String key = metric.substring(0, metric.lastIndexOf("="));
+        String val = metric.substring(metric.lastIndexOf("=") + 1);
+        metrics.put(key, val);
+      }
+    }
+    try {
+      LOG.info("Kafka Metrics Reporter sending metrics = " + _parser.writerWithDefaultPrettyPrinter().writeValueAsString(metrics));
```
lindong28: Would it be better to make this a debug level logging? If user wants to see metrics value in console, maybe they can use DefaultMetricsReporterService?

src/main/java/com/linkedin/kmf/services/configs/KafkaMetricsReporterServiceConfig.java
```diff
@@ -0,0 +1,77 @@
...
+  public static final String ZOOKEEPER_CONNECT_CONFIG = CommonServiceConfig.ZOOKEEPER_CONNECT_CONFIG;
+  public static final String ZOOKEEPER_CONNECT_DOC = CommonServiceConfig.ZOOKEEPER_CONNECT_DOC;
+
+  public static final String BOOTSTRAP_SERVERS_CONFIG = CommonServiceConfig.BOOTSTRAP_SERVERS_CONFIG;
+  public static final String BOOTSTRAP_SERVERS_DOC = CommonServiceConfig.BOOTSTRAP_SERVERS_DOC;
+
+  public static final String TOPIC_CONFIG = CommonServiceConfig.TOPIC_CONFIG;
+  public static final String TOPIC_DOC = CommonServiceConfig.TOPIC_DOC;
+
+  public static final String TOPIC_REPLICATION_FACTOR = "reporter.kafka.topic.replication.factor";
```
lindong28: Just one last minor comment. Since this service uses the config `report.interval.sec`, can we name this config as `reporter.kafka.topic.replication.factor` for consistency?

hackerwin7:
Thanks a lot! I have modified above.

src/main/java/com/linkedin/kmf/services/configs/KafkaMetricsReporterServiceConfig.java
hackerwin7: Okay, I'll update to `report.kafka.topic.replication.factor`

lindong28:
@hackerwin7 Thanks for the patch! I will merge it.

PR #110, opened 2018-01-24
Config refactor to Typesafe Config

SlevinBE:
This PR is a refactor of the configuration mechanism of the Kafka Monitor from json/map to [Typesafe Config](https://github.com/lightbend/config).

**Why?**
* I needed to override several configuration values with environment variables (I'm using it within Docker). Typesafe Config makes this much easier with its ${?MY_ENV_VAR} syntax.
* less verbose configuration syntax
* A more structured type to indicate the configuration (Config instead of Map). Also no more casting of values when retrieving a property, Typesafe Config has getString, getBoolean, getList etc. methods


lindong28:
Thanks for the patch @SlevinBE! I will review the patch later.

smccauliff:
What is the migration path for existing users?

lindong28:
Thanks for the patch @SlevinBE. I think it definitely makes sense to be able to reference environment variable in the configuration. I just wondering if it is possible to make the change backward compatible instead of forcing user to make changes immediately.

This patch seems to break the existing Kafka Monitor usage in two ways.The first issue is that `./bin/kafka-monitor-start.sh config/kafka-monitor.properties` will not work unless user changes `kafka-monitor.properties` file. I think maybe we can change the KafkaMonitor.java in such a way that it supports both old and new properties file. For example, maybe the code can try to parse the properties file with the new logic, and if an exception is thrown, it can parse the file with the old logic. We can still print warning so that user of the old properties file can make changes to use the new format.

The second issue is that some user may instantiate KafkaMonitor class directly with a key/value map. This patch changes the constructor to accept an Config object. Maybe we can still keep old constructor and deprecate it.

What do you think?

SlevinBE:
Thank you for the feedback, certainly valid remarks.
HOCON, the format used by Typesafe Config, is a superset of JSON, so it might be possible to make it 100% backwards compatible. I know Typesafe Config can load pure json files, but I'll have to see how it handles this one (given this one has comments in it, which isn't pure json).

It is also possible to load a configuration in Typesafe Config from a key/value map, so I can probably also add the old constructor back in.

I'll try to get these changes in.

smccauliff:
Does typesafe config have dependencies on scala?  kafka-monitoring does have dependencies on scala because of the kafka server jar, but this may go away if we adopt use of the new Java AdminClient.

SlevinBE:
No, Typesafe Config is a Java library in the purest sense of the word. Its only dependency is the Java SDK and doesn't depend on any other libraries.

src/main/java/com/linkedin/kmf/apps/MultiClusterMonitor.java
```diff
@@ -34,38 +35,34 @@
...
 
-  public MultiClusterMonitor(Map<String, Object> props, String name) throws Exception {
+  public MultiClusterMonitor(Config serviceConfig, String name) throws Exception {
     _name = name;
-    MultiClusterMonitorConfig config = new MultiClusterMonitorConfig(props);
-    _topicManagementService = new MultiClusterTopicManagementService(createMultiClusterTopicManagementServiceProps(props, config), name);
-    _produceService = new ProduceService(createProduceServiceProps(props, config), name);
-    _consumeService = new ConsumeService(createConsumeServiceProps(props, config), name);
+    MultiClusterMonitorConfig config = new MultiClusterMonitorConfig(serviceConfig);
+    _topicManagementService = new MultiClusterTopicManagementService(createMultiClusterTopicManagementServiceProps(serviceConfig, config), name);
```
andrewchoi5: since there are new commits to the _master_ branch, let's **resolve** the **conflicts** for better review.

SlevinBE:
closing this PR, as I think it's better to approach this a different way.

PR #111, opened 2018-01-25
upgrade to Kafka 0.11.0.1 to resolve percentile metrics issue (KAFKA-5203)

SlevinBE:
There's currently an issue with percentiles (such as records-delay-ms-99th) not being calculated correctly. For us, the effect was that the records-delay-ms-99th metric was gradually going down to 1 or 0ms and staying there forever.

This is being caused by an issue in Kafka 0.10.x, which is fixed in Kafka 0.11. More info on this issue can be found at https://github.com/apache/kafka/pull/3002

This PR upgrades the Kafka dependencies of Kafka Monitor to 0.11.0.1 to resolve this issue. The records-delay-ms-99th metric is now being shown correctly in our deployment of Kafka Monitor.
This PR has been tested on a Kafka 0.10.2.1 cluster.


SlevinBE:
Oops, opened the PR without having a look at the other open PRs. Turns out there was already one for the Kafka 0.11 upgrade. 
Anyway, the description of this PR might still have some value as an argument for the upgrade.

lindong28:
Thanks for the patch @SlevinBE. I have updated kafka-monitor to use 0.11 Kafka.

PR #114, opened 2018-03-09
Configurable Jolokia port/uri in JolokiaService and JettyService

### LINKED ISSUE #113: Make Jolokia port/url configurable ( in JolokiaService and in webapp/index.html )
Hi,

i had a requirement to deploy kafka-monitor in kubernetes, and expose JettyService ( specifically, the small js ui that draws a couple graphs ) and JolokiaService over an ingress.

In the current master branch this is rather difficult because index.html hardcodes the jolokia port/url as http:// + location.host.split(':')[0] + :8778/jolokia , with the port also hardcoded in JolokiaService.

I will be sending a PR for your review to discuss/potentially fix these issues.

thanks!
Andras

andlaz:
fixes #113 

it's been a while since i last touched JAX-RS, jetty, jersey or javascript :-) ..so this is a work in progress

### DIFF
--- a/src/main/java/com/linkedin/kmf/services/JettyService.java
+++ b/src/main/java/com/linkedin/kmf/services/JettyService.java
@@ -9,9 +9,18 @@
  */
 package com.linkedin.kmf.services;
 
+import com.fasterxml.jackson.jaxrs.json.JacksonJsonProvider;
+import com.linkedin.kmf.common.ConfigPublisher;
 import com.linkedin.kmf.services.configs.JettyServiceConfig;
+import com.linkedin.kmf.services.configs.JolokiaServiceConfig;
+import org.eclipse.jetty.server.Handler;
 import org.eclipse.jetty.server.Server;
+import org.eclipse.jetty.server.handler.HandlerCollection;
 import org.eclipse.jetty.server.handler.ResourceHandler;
+import org.eclipse.jetty.servlet.ServletContextHandler;
+import org.eclipse.jetty.servlet.ServletHolder;
+import org.glassfish.jersey.server.ResourceConfig;
+import org.glassfish.jersey.servlet.ServletContainer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/linkedin/kmf/services/JettyService.java
```diff
@@ -25,16 +33,39 @@
   private final Server _jettyServer;
   private final int _port;
 
+  private class ConfigPublisherApp extends ResourceConfig {
```
lindong28: Would it be more intuitive (and possible) to publish the jolokia port as part of the jolokia config rather than the jetty config?

webapp/index.html
```diff
@@ -156,7 +156,7 @@ <h1 style="text-align:center">Kafka Monitor GUI</h1>
     return charts;
   }
 
-  function updateChartAndValueTable(mbeans, charts, cells) {
+  function updateChartAndValueTable(j4p, mbeans, charts, cells) {
```
lindong28: Do we need the j4p parameter in this method?

src/main/java/com/linkedin/kmf/services/configs/JettyServiceConfig.java
```diff
@@ -20,12 +20,20 @@
...
+  public static final String JOLOKIA_URL_DOC = "URL to the Jolokia endpoint";
+
   static {
     CONFIG = new ConfigDef().define(PORT_CONFIG,
                                     ConfigDef.Type.INT,
                                     8000,
                                     ConfigDef.Importance.LOW,
-                                    PORT_DOC);
+                                    PORT_DOC)
+                            .define(JOLOKIA_URL_CONFIG,
```
lindong28: It seems a bit weird to have the number `8778` defined in both JettyServiceConfig and the JolokiaServiceConfig. Can we only define and use this in JolokiaServiceConfig?

src/main/java/com/linkedin/kmf/services/JettyService.java
andlaz: My problem with that is the implicit dependency that would create between JolokiaService and JettyService. Given how services are initialized and configured, how do you see JettyService loading the configuration of JolokiaService?

webapp/index.html
andlaz: j4p would no longer be in scope for this method otherwise. If you have a more elegant solution, please let me know ( i'm sure there is .. )

src/main/java/com/linkedin/kmf/services/configs/JettyServiceConfig.java
andlaz: Please see my comment above for the dependency this would create between the two services.

andlaz:
No problem, thanks for the review @lindong28 !

webapp/index.html
lindong28: Ah I didn't notice that j4p is already used in the function. Sure, it makes sense.

src/main/java/com/linkedin/kmf/services/configs/JettyServiceConfig.java
andlaz: I'm not sure leaving the end user to carry this configuration detail in query params ( a bookmark or something similar ) would be the best. At minimum, it would be initially confusing.

I prefer "just works" :-)

lindong28: @andlaz I also prefer things to just work. But user already needs to explicitly specify the jetty port (e.g. 8000) anyway. So I am not sure how much different it makes to the user to not having to specify the jolokia port. We can just provide user a link (e.g. localhost:8000/index.html?jolokiaPort=8779&joloiaIP=localhost) that they can use to access the web GUI rather than giving them a port number and ask them figure out the link. And under this assumption, then it does not really matter whether there is one more parameter in the link. What do you think?

I am trying to see whether the additional code complexity is worth the convenience it gives to the user.

andrewchoi5:
If this is an inactive Pull Request, may I have your permission to close it or update?

src/main/java/com/linkedin/kmf/services/JettyService.java
```diff
@@ -9,9 +9,17 @@
  */
```
andrewchoi5: If this is an inactive `Pull Request`, may I have your permission to close it or update?
 😄 
Thanks!

andrewchoi5: @andlaz

andlaz: feel free to update or close @andrewchoi5

PR #122, opened 2018-05-23
Add 99th and 999th percentile produce delay metrics

lindong28:


lindong28:
@smccauliff Could you review this patch? Thanks!

src/main/java/com/linkedin/kmf/services/ProduceService.java
```diff
@@ -77,14 +79,17 @@
...
+  private final int _latencyPercentileGranularityMs;
 
   public ProduceService(Map<String, Object> props, String name) throws Exception {
     _name = name;
     ProduceServiceConfig config = new ProduceServiceConfig(props);
     _zkConnect = config.getString(ProduceServiceConfig.ZOOKEEPER_CONNECT_CONFIG);
     _brokerList = config.getString(ProduceServiceConfig.BOOTSTRAP_SERVERS_CONFIG);
     String producerClass = config.getString(ProduceServiceConfig.PRODUCER_CLASS_CONFIG);
-
+    _latencyPercentileMaxMs = config.getInt(ProduceServiceConfig.LATENCY_PERCENTILE_MAX_MS_CONFIG);
```
smccauliff: It's not clear what this is supposed to be.  By reading the code it looks like percentiles are based on some fixed sized histogram and this is the maximum value allowed by the histogram.   Perhaps maxLatencyHistogram or maxBinnedValue or something like that would be better.

src/main/java/com/linkedin/kmf/services/ProduceService.java
```diff
@@ -77,14 +79,17 @@
...
+    _latencyPercentileGranularityMs = config.getInt(ProduceServiceConfig.LATENCY_PERCENTILE_GRANULARITY_MS_CONFIG);
```
smccauliff: Same type of comment here.  Histogram bin size or bucket width or interval size seems more clear to me.

src/main/java/com/linkedin/kmf/services/ProduceService.java
```diff
@@ -241,6 +246,13 @@ public ProduceMetrics(final Metrics metrics, final Map<String, String> tags) {
       _produceDelay.add(new MetricName("produce-delay-ms-avg", METRIC_GROUP_NAME, "The average delay in ms for produce request", tags), new Avg());
       _produceDelay.add(new MetricName("produce-delay-ms-max", METRIC_GROUP_NAME, "The maximum delay in ms for produce request", tags), new Max());
 
+      // There are 2 extra buckets use for values smaller than 0.0 or larger than max, respectively.
+      int bucketNum = _latencyPercentileMaxMs / _latencyPercentileGranularityMs + 2;
+      int sizeInBytes = 4 * bucketNum;
+      _produceDelay.add(new Percentiles(sizeInBytes, _latencyPercentileMaxMs, Percentiles.BucketSizing.CONSTANT,
```
smccauliff: Using fixed width histograms  seems like a poor choice for capturing the tails of distributions.  Kafka  does not have anything else to choose from?  Exponential decay?  Dynamic histograms?

lindong28: Yeah Kafka does not have other choice.

src/main/java/com/linkedin/kmf/services/ProduceService.java
lindong28: Thanks for the review!

Yes the implementation of this metric is based on bucket. On the other hand the metric exposed to user is actually a percentile number rather than histogram. So it maybe better to have `percentile` in the metric name to refer to the interpretation of the metric rather than using `histogram` to refer to the implementation of this metric.

Also, this config is very similar to the existing config `consume.latency.percentile.max.ms` in the ConsumeServiceConfig. So it maybe better to follow the same pattern for consistency?

lindong28:
Thanks for the review @sutambe

lindong28:
Thanks for the review @smccauliff

PR #123, opened 2018-05-25
Fixes double appending of prefix in metrics name

jlisam:
The current statsd client already appends the prefix at the beginning of the metric name. Unfortunately the current code was also appending the prefix. For example:

If `prefix` was `kafka-monitor`, the final metric name produced would be:
`kafka-monitor.kafka-monitor.produce_service.single-cluster-monitor.produce-availability-avg`

But in reality should be:
`kafka-monitor.produce_service.single-cluster-monitor.produce-availability-avg`

This PR solves that problem by letting the statsd client do the appending of the prefix and removing some of that logic in the `StatsdMetricsReporterService`

### DIFF
--- a/src/main/java/com/linkedin/kmf/services/StatsdMetricsReporterService.java
+++ b/src/main/java/com/linkedin/kmf/services/StatsdMetricsReporterService.java
@@ -34,7 +34,6 @@ public class StatsdMetricsReporterService implements Service {
   private final int _reportIntervalSec;
   private final ScheduledExecutorService _executor;
   private final StatsDClient _statsdClient;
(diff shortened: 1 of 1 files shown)


lindong28:
Thanks for the patch @jlisam! This patch was not developed or used at LinkedIn. It maybe better for the original author to comment on this.

Hey @arsenaid, would you have time to review this patch?

lindong28:
@jlisam Thanks for the patch! I am not able to review or test this patch because LinkedIn does not use StatsdMetricsReporterService. I will just merge this PR.

jlisam:
thanks @lindong28 for the support!

PR #125, opened 2018-06-13
Fix misspelling

jonathansantilli:


lindong28:
Thanks for the fix @jonathansantilli. Sorry for the delay. LGTM.

PR #128, opened 2018-07-12
Add support for reporting metrics to SignalFx

rohitvyavahare:
1. Adds a metrics reporting service to report the Kafka monitor metrics to SignalFx.
2. Suppress checkstyle of LinkedIn Corp. headers for SignalFx files
3. Add avg-delay metric per broker to determine latency for each broker from producer and consumer

SignalFx is monitoring platform that performs operational intelligence for data-driven DevOps

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -55,6 +55,7 @@ allprojects {
 
   checkstyle {
     configFile = new File(rootDir, "checkstyle/checkstyle.xml")
+    configProperties = [ "suppressionFile" : new File(rootDir, "checkstyle/suppressions.xml") ]
   }
 
   test.dependsOn('checkstyleMain', 'checkstyleTest')
--- a/build.gradle
+++ b/build.gradle
@@ -25,6 +25,7 @@ allprojects {
     compile 'org.jolokia:jolokia-jvm:1.3.3'
     compile 'net.savantly:graphite-client:1.1.0-RELEASE'
     compile 'com.timgroup:java-statsd-client:3.0.1'
+    compile 'com.signalfx.public:signalfx-codahale:0.0.47'
 
     testCompile 'org.testng:testng:6.8.8'
   }
--- a/checkstyle/checkstyle.xml
+++ b/checkstyle/checkstyle.xml
@@ -21,6 +21,10 @@
     <property name="header" value="/\*\*\nCopyright 2016 LinkedIn Corp. Licensed under the Apache License.*"/>
   </module>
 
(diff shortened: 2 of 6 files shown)


lindong28:
Hey @rohitvyavahare, I just got reply from open source team at LinkedIn. It looks like we would liek you to confirm that you can license this patch to us under Apache 2 license. Would you be OK to delete “All rights reserved” and add “Licensed under the Apache 2 License”?

Ping @MBShannon for more comments.

lindong28:
@rohitvyavahare Another question here is that could you confirm this code is written by you? The open source team would like to double check.

lindong28:
@rohitvyavahare Another question here is that could you confirm this code is written by you as part of SignalFx? The open source team would like to double check.

rohitvyavahare:
@lindong28 Thank you for getting back to us. Yes, I am fine with changing "All rights reserved" to "Licensed under the Apache 2 License". For the second question, what form of confirmation you need?

lindong28:
@rohitvyavahare Thanks! For the second question, I guess it just needs verbal confirmation. @MBShannon asked about this.

rohitvyavahare:
@lindong28  @MBShannon I confirm that I have written above code on behalf of SignalFx, Inc. Also, I have updated files with **Licensed under the Apache 2 License**

lindong28:
@rohitvyavahare Thanks for the update. It seems that the patch is updated on July 30 to include the average per-broker latency. In general Kafka Monitor attempts to monitor Kafka cluster performance from client (i.e. producer and consumer) point of view and those users probably should not have to know urls of each broker in the cluster. I can see the value of monitoring per-broker latency for system administrator. On the other hand today our system administrator can already monitor this per-broker performance by watching e.g. `kafka.network:type=RequestMetrics,name=LocalTimeMs,request={Produce|FetchConsumer|FetchFollower}` which is available the Kafka server metrics.

So it seems that the newly added metrics kind of deviate from kafka-monitor's intended use-case and its functionality overlaps with the existing metrics in Kafka server. I am not sure we need this. Since main goal of this patch is to export the existing kafka-monitor metrics to SignalFx, can we remove this part from this patch and put it in a separate PR if needed?

rohitvyavahare:
@lindong28 I see your point. I have updated PR with only SignalFx metric reporter. I will create new PR for broker metrics. We can discuss more in that PR. Thanks

lindong28:
Thanks for the patch @rohitvyavahare. Merged to master branch.

PR #131, opened 2018-08-16
Upgrade to kafka 1.1.1

lindong28:


lindong28:
@smccauliff This patch upgrades kafka-monitor to 1.1.1. I have tested patch locally. Can you take a look when you get time?

lindong28:
@smccauliff I added the following comment `Need to catch throwable because there is scala API that can throw NoSuchMethodError in runtime and such error is not caught by compilation`.

smccauliff:
Thanks for doing this.

PR #153, opened 2019-04-02
add bintray suppport and upgrade gradle

xiowu0:
add bintray support


build.gradle
```diff
@@ -1,14 +1,29 @@
...
+apply plugin: 'com.jfrog.bintray'
+
 allprojects {
   apply plugin: 'idea'
   apply plugin: 'eclipse'
   apply plugin: 'java'
   apply plugin: 'checkstyle'
 
-  version = "1.0.0"
+  version = "2.0.0"
```
radai-rosenblatt: if we're usingthe semantic build versioning plugin than we dont want to hardcode a version

xiowu0: removed hard coded version

PR #154, opened 2019-05-07
allow brokers to be blacklisted when adding new partitions or reassigning partitions

xiowu0:
KMF today assign partitions or add new partitions to all kafka brokers in a cluster. However, there are use cases that we don't want KMF topic partitions to be created on certain brokers. For example, broker that is not in a healthy state and is being removed might not want to take any new KMF partitions.  This fix allows user to blacklist certain kafka brokers via topic factory so that blacklisted brokers will not be assigned new partitions or taking reassigned partitions. The default topic factory implementation doesn't blacklist any brokers. User can have custom implementation of topic factory where blacklisted broker info can be obtained dynamically.

### DIFF
--- a/src/main/java/com/linkedin/kmf/services/MultiClusterTopicManagementService.java
+++ b/src/main/java/com/linkedin/kmf/services/MultiClusterTopicManagementService.java
@@ -265,14 +265,34 @@ public class MultiClusterTopicManagementService implements Service {
         if (partitionNum < minPartitionNum) {
           LOG.info("MultiClusterTopicManagementService will increase partition of the topic {} "
               + "in cluster {} from {} to {}.", _topic, _zkConnect, partitionNum, minPartitionNum);
-
+          Collection<Integer> blackListedBrokers =
+              _topicFactory.getBlackListedBrokers(_zkConnect);
           scala.Option<scala.collection.Map<java.lang.Object, scala.collection.Seq<java.lang.Object>>> replicaAssignment = scala.Option.apply(null);
           scala.Option<Seq<Object>> brokerList = scala.Option.apply(null);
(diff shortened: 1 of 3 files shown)


src/main/java/com/linkedin/kmf/topicfactory/TopicFactory.java
```diff
@@ -34,4 +35,10 @@
 
   int createTopicIfNotExist(String zkUrl, String topic, int replicationFactor, double partitionToBrokerRatio, Properties topicProperties);
 
+  /**
+   * @param zkUrl zookeeper connection url
+   * @return A list of brokers that don't take new partitions or reassigned partitions for topics.
+   */
+  Collection<Integer> getBlackListedBrokers(String zkUrl);
```
smccauliff: Set<Integer>

src/main/java/com/linkedin/kmf/services/MultiClusterTopicManagementService.java
```diff
@@ -265,23 +265,43 @@ void maybeAddPartitions(int minPartitionNum) {
...
+          adminZkClient.addPartitions(_topic, existingAssignment,
+              scala.collection.JavaConversions.collectionAsScalaIterable(brokers).toSeq(),
+              minPartitionNum, replicaAssignment, false);
         }
       } finally {
         zkClient.close();
       }
     }
 
+    private Collection<Broker> getAvailableBrokers(KafkaZkClient zkClient) {
```
smccauliff: Set<Broker> don't want to list a broker more than once.

src/main/java/com/linkedin/kmf/topicfactory/TopicFactory.java
xiowu0: Make sense. fixes.

src/main/java/com/linkedin/kmf/services/MultiClusterTopicManagementService.java
xiowu0: fixed

PR #162, opened 2019-09-13
Add a build step for docker container

levonet:


smccauliff:
We don't use docker much a LinkedIn.  Is it obvious to a user of docker how to use this functionality? Otherwise can you update the README.md with instructions on how to build the docker image.

levonet:
Nothing will change for users.
This change makes a build of an app before building container. It's needed for automatic container building without user environment.
`latest` is a default version name of docker containers.

levonet:
@smccauliff ^^^

docker/Dockerfile
```diff
@@ -10,18 +10,26 @@
...
+ADD . .
+RUN ./gradlew jar
 
 FROM anapsix/alpine-java
 
 MAINTAINER coffeepac@gmail.com
 
 WORKDIR /opt/kafka-monitor
-ADD build/ build/
+COPY --from=build /home/gradle/project/build build/
```
smccauliff: Should't this use relative paths rather than absolute paths?

docker/Makefile
```diff
@@ -14,12 +14,11 @@
 
 all: push
 
-TAG ?= 0.1
+TAG ?= latest
```
smccauliff: Wouldn't it be better to get the version from gradle?

andrewchoi5:
@levonet If this is an inactive Pull Request, may I have your permission to close it?

docker/README.md
```diff
@@ -7,7 +7,7 @@ build will error out complaining of missing files.
 
 ### Build the container locally
 Decide where you would like the built container to be stored.  The
```
andrewchoi5: If this is an inactive `Pull Request`, may I have your permission to close it or update?
 😄 
Thanks!

andrewchoi5: @levonet

levonet: Yes. You can do it on your own.

docker/Dockerfile
levonet: The line above is set to the working folder `WORKDIR`.
Relative paths are used below this change.
Change everything or leave the consistency?

docker/Makefile
levonet: Possible, but in a Makefile this can be passed through an environment variable.
`TAG=${VERSION} make`

It is better to do this in CI than in a Makefile.

andrewchoi5:
Closing due to inactivity.

PR #163, opened 2019-11-06
Addition of Tracelogs for recordsLost in Consume Service 

andrewchoi5:
Addition of Tracelogs for **recordsLost** in _Consume Service_ to indicate record index and the next index.


src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -173,7 +173,9 @@ private void consume() throws Exception {
         _sensors._recordsDuplicated.record();
       } else if (index > nextIndex) {
         nextIndexes.put(partition, index + 1);
-        _sensors._recordsLost.record(index - nextIndex);
+        double numLostRecords = index - nextIndex;
```
smccauliff: Why double?

andrewchoi5: amended to _long_

PR #174, opened 2019-12-06
Single Cluster Monitoring: ensure there are no static imports.

andrewchoi5:
**Single Cluster** Monitoring - ensure there are no _static_ imports.


andrewchoi5:
> The change definitely makes sense. But can you briefly explain why is this change a good idea? Is it just a style issue fix, eliminates potential confusion, or increase performance?

This pertains to readability. Most developers know that most lines of code is read and examined much more than it is originally written. Thus, being able to comprehend the lines of source code is much more critical than the swift writability. Static import can be not so good since static imports add an exception to the rules which every developer has to remember.

PR #175, opened 2019-12-06
Migration from ZkClient to AdminClient for KMF Utils Class

andrewchoi5:
Migration from ZkClient to AdminClient for KMF Utils Class.

Testing: ran the tests locally on local machine with 1, 3, and 5 kafka-servers / brokers and used kafka-topic.sh tool to increase the partition counts. I manipulated the number of partitions for the topic, and there was not any exceptions occurred or illegal shutdowns.


andrewchoi5:
> Good work. Just some thoughts.
> 
> I think the reason why you migrated from ZkClient to AdminClient is mostly because `ZkUtils` class has been deprecated since `2.0.0`. The `AdminClient` basically sends a metadata request to Kafka server and get the most up-to-date metadata (hopefully). However, in Kafka, metadata is eventually consistent. That means `AdminClient` might retrieve some outdated metadata that can not accurately describe the current state of cluster at the current movement. Maybe `AdminClient` has some special handling of this case. But I am not sure. I am wondering that have you considered this edge case and what potentially risk it might impose to Kafka monitor?
> 
> In some other project when I need to get cluster description, I usually read these information directly off Zookeeper which is the real source of truth (maybe I need to have more trust on `AdminClinet` and take advantage of its convenience).
> 
> So I am just wondering what is your thoughts on it.

Thank you for raising your opinion. 
The Future.get() returns only after successful computation. Future.get() documentation. 

> Waits if necessary, for the computation to complete and then retrieves its result.

When the `Future's` `get()` method gets called, it throws **ExecutionException**, which the caller of `get()` then unpacks it and handles. Normal exception handling is caused into the Callable or Future pattern model.

src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
```diff
@@ -49,7 +49,7 @@
   private final ConsumeService _consumeService;
   private final String _name;
 
-  public SingleClusterMonitor(Map<String, Object> props, String name) throws Exception {
+  private SingleClusterMonitor(Map<String, Object> props, String name) throws Exception {
```
smccauliff: Since this class implements App changing the  visibility of this method  may break other users code.

src/main/java/com/linkedin/kmf/services/ProduceService.java
```diff
@@ -238,7 +238,7 @@ public boolean isRunning() {
     private final ConcurrentMap<Integer, Boolean> _produceErrorInLastSendPerPartition;
     private final Map<String, String> _tags;
 
-    public ProduceMetrics(final Metrics metrics, final Map<String, String> tags) {
+    ProduceMetrics(final Metrics metrics, final Map<String, String> tags) {
```
smccauliff: Since ProduceService implements App external users may have code that instantiates it.

src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
andrewchoi5: Resolved.

src/main/java/com/linkedin/kmf/common/Utils.java
```diff
@@ -52,20 +47,21 @@
...
    */
-  public static int getPartitionNumForTopic(String zkUrl, String topic) {
-    ZkUtils zkUtils = ZkUtils.apply(zkUrl, ZK_SESSION_TIMEOUT_MS, ZK_CONNECTION_TIMEOUT_MS, JaasUtils.isZkSecurityEnabled());
+  public static int getPartitionNumForTopic(AdminClient adminClient, String topic)
+      throws ExecutionException, InterruptedException {
     try {
-      Seq<String> topics = scala.collection.JavaConversions.asScalaBuffer(Arrays.asList(topic));
-      return zkUtils.getPartitionsForTopics(topics).apply(topic).size();
+      ArrayList<String> topics = new ArrayList<>();
+      topics.add(topic);
```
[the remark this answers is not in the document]
andrewchoi5: Collections singleton utilized.

src/main/java/com/linkedin/kmf/services/ProduceService.java
andrewchoi5: Amended for external clients.

PR #176, opened 2019-12-07
Executor Fixed rated schedulings Runnable -> lamda conversions

andrewchoi5:
Executor fixed rated schedulings lamda conversions from Runnables for services

### DIFF
--- a/src/main/java/com/linkedin/kmf/KafkaMonitor.java
+++ b/src/main/java/com/linkedin/kmf/KafkaMonitor.java
@@ -101,17 +101,13 @@ public class KafkaMonitor {
       entry.getValue().start();
     }
 
-    _executor.scheduleAtFixedRate(
-      new Runnable() {
-        @Override
-        public void run() {
-          try {
-            checkHealth();
-          } catch (Exception e) {
-            LOG.error("Failed to check health of tests and services", e);
-          }
-        }
-      }, 5, 5, TimeUnit.SECONDS
+    _executor.scheduleAtFixedRate(() -> {
+      try {
+        checkHealth();
+      } catch (Exception e) {
+        LOG.error("Failed to check health of tests and services", e);
+      }
+    }, 5, 5, TimeUnit.SECONDS
     );
   }
 
--- a/src/main/java/com/linkedin/kmf/KafkaMonitor.java
+++ b/src/main/java/com/linkedin/kmf/KafkaMonitor.java
@@ -101,13 +101,16 @@ public class KafkaMonitor {
       entry.getValue().start();
     }
 
+    long initialDelaySecond = 5;
+    long periodSecond = 5;
+
     _executor.scheduleAtFixedRate(() -> {
       try {
         checkHealth();
       } catch (Exception e) {
         LOG.error("Failed to check health of tests and services", e);
       }
-    }, 5, 5, TimeUnit.SECONDS
+    }, initialDelaySecond, periodSecond, TimeUnit.SECONDS
     );
   }
 
--- a/src/main/java/com/linkedin/kmf/common/DefaultTopicSchema.java
+++ b/src/main/java/com/linkedin/kmf/common/DefaultTopicSchema.java
@@ -15,17 +15,17 @@ import org.apache.avro.Schema.Field;
 
 public class DefaultTopicSchema {
 
-  public static final Field TOPIC_FIELD = new Field("topic", Schema.create(Schema.Type.STRING), null, null);
+  static final Field TOPIC_FIELD = new Field("topic", Schema.create(Schema.Type.STRING), null, null);
 
   public static final Field TIME_FIELD = new Field("time", Schema.create(Schema.Type.LONG), null, null);
 
   public static final Field INDEX_FIELD = new Field("index", Schema.create(Schema.Type.LONG), null, null);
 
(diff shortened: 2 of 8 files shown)


src/main/java/com/linkedin/kmf/KafkaMonitor.java
```diff
@@ -101,17 +101,13 @@ public synchronized void start() {
...
-          }
-        }
-      }, 5, 5, TimeUnit.SECONDS
+    _executor.scheduleAtFixedRate(() -> {
+      try {
+        checkHealth();
+      } catch (Exception e) {
+        LOG.error("Failed to check health of tests and services", e);
+      }
+    }, 5, 5, TimeUnit.SECONDS
```
Lincong: Nit: can you make these two variables for these two "5" ? I think it would increase the readability by a little
```java
long initialDelaySecond = 5;
long periodSecond = 5;

_executor.scheduleAtFixedRate(() -> {
       try {
         checkHealth();
       } catch (Exception e) {
         LOG.error("Failed to check health of tests and services", e);
       }
     }, initialDelaySecond, periodSecond, TimeUnit.SECONDS
```

src/main/java/com/linkedin/kmf/common/DefaultTopicSchema.java
```diff
@@ -15,17 +15,17 @@
 
 public class DefaultTopicSchema {
 
-  public static final Field TOPIC_FIELD = new Field("topic", Schema.create(Schema.Type.STRING), null, null);
+  static final Field TOPIC_FIELD = new Field("topic", Schema.create(Schema.Type.STRING), null, null);
```
Lincong: Nit: Reducing these unnecessary access scope (surface area of a class) is definitely a good thing to do in general. But on the other hand, this class is defined in the `common` package and it implies that this class is likely or supposed to be used in a shared way. So making some of the static functions package private just because they are not used at this point seems a little unnecessary for me. It's just my opinion (you don't have to change anything). But I'd like to know your thoughts about it.

src/main/java/com/linkedin/kmf/services/StatsdMetricsReporterService.java
```diff
@@ -47,17 +47,13 @@ public StatsdMetricsReporterService(Map<String, Object> props, String name) {
...
-        @Override
-        public void run() {
-          try {
-            reportMetrics();
-          } catch (Exception e) {
-            LOG.error(_name + "/StatsdMetricsReporterService failed to report metrics", e);
-          }
-        }
-      }, _reportIntervalSec, _reportIntervalSec, TimeUnit.SECONDS
+    _executor.scheduleAtFixedRate(() -> {
```
Lincong: I noticed that usage of the executor service is kind of prevalent in kafka-monitor. Instead of using the `ScheduledExecutorService` explicitly, would it be a good idea to write a thin wrapper/abstraction which takes in an executor and provided an unified API/method to schedule periodical task ? For example

```java
  class KmfExecutor {
    // ...
    public KmfExecutor(ScheduledExecutorService executor) {
      // ...
    }

    public void scheduleAtFixedRate(/* some function */, long initialDelay, long period, String errorMsg) {
      // ...
    }

    public void awaitTermination() {
      // ...
    }

    public void shutdown() {
      // ...
    }
    // ...
  }
```

Then instantiate them all together and provide each component that needs executor with this `KmfExecutor` so that each component do not have to create its own ad-hoc executor. Here are a few advantages:

1. Increase testability. We can potentially mock the `KmfExecutor` or make the `KmfExecutor` takes in mocked executor.

2. We can potentially have a centralized class which contains all executors and add metrics to this "executor manager"

It is just my thoughts and just want to let you know in case you think this is a good idea or potentially something you might want to do. You don't have to make any change at this point.

src/main/java/com/linkedin/kmf/common/DefaultTopicSchema.java
andrewchoi5: You've raised a reasonable point. Decreasing the surface area of a class is a sound thing to perform. Restricting the avenues which classes can interact with this one makes it simpler to render internal modifications that are not visible to those other classes. We need to expose only to **bare minimum** which is required to fulfill these requirements.

`DefaultTopicSchema` is in fact defined within the `common` package. However public `surface area` for a class effectively defines its `API`, and we can expose it back to public if it should be accessible to wider scope in the future. 

If variables are exposed via methods, the logistics of how they work can be modified later.

When exposing  a field, some other class outside of control and maybe outside of **knowledge** beings referencing such field -> leading to being stuck with exposing that field for the rest of time. This might lead up to breaking `backwards-compatibility`.

In general, not  "all classes" should have access them. In pursuit of **maintainability** of source code, we shouldn't necessarily go for the most convenient to write in the first place.

src/main/java/com/linkedin/kmf/services/StatsdMetricsReporterService.java
andrewchoi5: Composing a `wrapper or abstraction` which takes in an executor and provides a **unified API** and method to _schedule periodical task_ sounds like a reasonable idea.

I am also for avoiding individually-owned ad-hoc executor. I can do this in the future in spare time for the benefits of increased `testability` and `class centralization`.

src/main/java/com/linkedin/kmf/KafkaMonitor.java
andrewchoi5: Amended and resolved.

PR #179, opened 2019-12-10
Remove Static Import & Allow conciseness on Measurables using lamda expression

andrewchoi5:
Allow conciseness on Measurables and Remove `Static Import`


src/main/java/com/linkedin/kmf/partitioner/NewKMPartitioner.java
```diff
@@ -9,13 +9,11 @@
...
 package com.linkedin.kmf.partitioner;
 
-import static org.apache.kafka.common.utils.Utils.murmur2;
-
 public class NewKMPartitioner implements KMPartitioner {
 
   public int partition(String key, int partitionNum) {
     byte[] keyBytes = key.getBytes();
-    return toPositive(murmur2(keyBytes)) % partitionNum;
+    return toPositive(org.apache.kafka.common.utils.Utils.murmur2(keyBytes)) % partitionNum;
```
smccauliff: Again, don't understand the war on static imports.

andrewchoi5: `Avoiding` _wildcard imports_ would be more preferable in general cases. I could let the other wildcard imports let it reside going forward.

PR #180, opened 2019-12-10
Remove ZooKeeper URL within Topic Factory Blacklisted Brokers 

andrewchoi5:
Remove **ZooKeeper** URL within Topic Factory `Blacklisted Brokers`.
Currently, the interfaces and the implementation takes in _zkUrl_ but it not used in the logic.


smccauliff:
The zkString in the method signature is still needed by LinkedIn  internal classes.  Making the atLeast() a regular import just makes everything  less readable.

andrewchoi5:
> The zkString in the method signature is still needed by LinkedIn internal classes. Making the atLeast() a regular import just makes everything less readable.

Static imports I believe should be used sparingly and rarely unless you would  
> otherwise be tempted to declare local copies of constants, or to abuse inheritance (the Constant Interface Antipattern)

https://stackoverflow.com/questions/420791/what-is-a-good-use-case-for-static-import-of-methods

However, _going forward after this PR, I will pass checking for the `static` imports._

smccauliff:
This is the intended use of atLeast()

See https://github.com/apache/kafka/blob/master/clients/src/main/java/org/apache/kafka/clients/consumer/ConsumerConfig.java

andrewchoi5:
> This is the intended use of atLeast(); to use it in a static import.
> 
> See https://github.com/apache/kafka/blob/master/clients/src/main/java/org/apache/kafka/clients/consumer/ConsumerConfig.java

Seems like they are declaring two static imports that are not wildcard imports.
`import static org.apache.kafka.common.config.ConfigDef.Range.atLeast`;
`import static org.apache.kafka.common.config.ConfigDef.ValidString.in`;

Sounds good.

PR #181, opened 2019-12-11
Correct the ordering of metricConfig configuration in Consume Service and make MetricConfig execute after topic is ready

andrewchoi5:
The original source code has the `consume()` thread starting running before the MetricConfig is configured. Thus, this PR corrects the `ordering` of **metricConfig** configuration within the `Consume Service`.
Additionally, this PR enables `MetricConfig` to execute after topic and its respective topic partitions are ready. Tested locally.


src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -105,25 +103,26 @@ public ConsumeService(Map<String, Object> props, String name) throws Exception {
...
+      reporters.add(new JmxReporter(JMX_PREFIX));
+      Metrics metrics = new Metrics(metricConfig, reporters, new SystemTime());
+      Map<String, String> tags = new HashMap<>();
+      tags.put("name", _name);
+      _adminClient = AdminClient.create(props);
+      _sensors = new ConsumeMetrics(metrics, tags, topic);
+      _thread = new Thread(() -> {
+        try {
+          consume();
+        } catch (Exception e) {
```
smccauliff: This approach, waiting for the topic to be ready, seems like it will fix some problems with order.  But making the consume service more resilient to error conditions in general will handle cases that has not yet been thought of.  When is it OK to exit the consume service?

src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -23,10 +23,16 @@
...
   private final MultiClusterTopicManagementService _multiClusterTopicManagementService;
+  private static final CompletableFuture<Void> TOPIC_PARTITION_READY = new CompletableFuture<>();
 
   public TopicManagementService(Map<String, Object> props, String serviceName) throws Exception {
     Map<String, Object> serviceProps = createMultiClusterTopicManagementServiceProps(props, serviceName);
     _multiClusterTopicManagementService = new MultiClusterTopicManagementService(serviceProps, serviceName);
+    TOPIC_PARTITION_READY.complete(null);
+  }
+
+  public static CompletableFuture<Void> topicPartitionReady() {
```
smccauliff: Static methods are going to kill any kind of testability which is why there's a suggestion to work on unit testing.  Then it will become more clear what's safe to make static.

src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -23,10 +23,16 @@
  */
 public class TopicManagementService implements Service {
   private final MultiClusterTopicManagementService _multiClusterTopicManagementService;
+  private static final CompletableFuture<Void> TOPIC_PARTITION_READY = new CompletableFuture<>();
```
smccauliff: What  happens if there is more than one topic management service in a process?

src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
```diff
@@ -43,7 +43,6 @@
 
 public class SingleClusterMonitor implements App {
   private static final Logger LOG = LoggerFactory.getLogger(SingleClusterMonitor.class);
-
```
smccauliff: Recommend adding the gradle checkstyle plugin to enforce the consistency that is desired here.  Then address all the checkstyle violations in a single PR.  After that have it fail the build on violations.

src/main/java/com/linkedin/kmf/services/TopicManagementService.java
andrewchoi5: updated.

src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
andrewchoi5: ready in a separate PR.

src/main/java/com/linkedin/kmf/services/TopicManagementService.java
andrewchoi5: Non-static usages now. Unit testing is in the backlog.

src/main/java/com/linkedin/kmf/services/ConsumeService.java
andrewchoi5: Updated the PR. Which other avenues to render consume service more resilient to error conditions do you suggest? Currently, with the combinations of `asynchronous` calls due to the nature of Kafka Monitor since its inception and the usages of the new `AdminClient` API, it is appropriate to align the executions.

config/multi-cluster-monitor.properties
```diff
@@ -12,12 +12,13 @@
 # each cluster in the pipeline. The "produce.service.props" should use the first cluster and
 # the "consume.service.props" should use the last cluster in the pipeline.
 
+# Produce service: Configure Produce Service to produce to the first cluster of the pipeline
```
andrewchoi5: comments `repositioned` to the top since it was interfering with the properties compilation.

src/main/java/com/linkedin/kmf/apps/MultiClusterMonitor.java
```diff
@@ -29,17 +30,18 @@
...
   private final ProduceService _produceService;
   private final ConsumeService _consumeService;
   private final String _name;
 
   public MultiClusterMonitor(Map<String, Object> props, String name) throws Exception {
     _name = name;
     MultiClusterMonitorConfig config = new MultiClusterMonitorConfig(props);
-    _topicManagementService = new MultiClusterTopicManagementService(createMultiClusterTopicManagementServiceProps(props, config), name);
+    _multiClusterTopicManagementService = new MultiClusterTopicManagementService(createMultiClusterTopicManagementServiceProps(props, config), name);
+    CompletableFuture<Void> topicPartitionReady = _multiClusterTopicManagementService.topicPartitionReady();
```
sutambe: method name and future name `topicPartitionReady` is a little confusing. This operation could fail so consider naming the method `startBlahBlahAsync` and the future `blahBlahResult`.

src/main/java/com/linkedin/kmf/apps/MultiClusterMonitor.java
```diff
@@ -29,17 +30,18 @@
 public class MultiClusterMonitor implements App {
   private static final Logger LOG = LoggerFactory.getLogger(MultiClusterMonitor.class);
 
-  private final MultiClusterTopicManagementService _topicManagementService;
+  private final MultiClusterTopicManagementService _multiClusterTopicManagementService;
   private final ProduceService _produceService;
   private final ConsumeService _consumeService;
   private final String _name;
 
   public MultiClusterMonitor(Map<String, Object> props, String name) throws Exception {
```
sutambe: I like to avoid kicking off async operations in the constructor. The main reason is that constructors can't return anything. I would strongly encourage you to separate construction from kicking off the async operation so that the future can be returned.

src/main/java/com/linkedin/kmf/apps/MultiClusterMonitor.java
```diff
@@ -29,17 +30,18 @@
...
     _name = name;
     MultiClusterMonitorConfig config = new MultiClusterMonitorConfig(props);
-    _topicManagementService = new MultiClusterTopicManagementService(createMultiClusterTopicManagementServiceProps(props, config), name);
+    _multiClusterTopicManagementService = new MultiClusterTopicManagementService(createMultiClusterTopicManagementServiceProps(props, config), name);
+    CompletableFuture<Void> topicPartitionReady = _multiClusterTopicManagementService.topicPartitionReady();
     _produceService = new ProduceService(createProduceServiceProps(props, config), name);
-    _consumeService = new ConsumeService(createConsumeServiceProps(props, config), name);
+    _consumeService = new ConsumeService(createConsumeServiceProps(props, config), name, topicPartitionReady);
```
sutambe: Why does `consumeService` accept the future but not `produceService`?

src/main/java/com/linkedin/kmf/apps/MultiClusterMonitor.java
```diff
@@ -70,28 +72,31 @@ public MultiClusterMonitor(Map<String, Object> props, String name) throws Except
...
-    _produceService.start();
-    _consumeService.start();
-    LOG.info(_name + "/MultiClusterMonitor started");
+    _multiClusterTopicManagementService.start();
+    CompletableFuture<Void> completableFuture = _multiClusterTopicManagementService.topicManagementReady();
+    completableFuture.thenRun(() -> {
+      _produceService.start();
+      _consumeService.start();
+    });
+    LOG.info(_name + "/MultiClusterMonitor started.");
```
sutambe: I would put the log line inside `.thenRun` lambda. Secondly, if `producerService.start` throws, should `consumerService` start? Perhaps not. Consider using `CompletableFuture.handle` to handle exceptions.

src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
```diff
@@ -52,8 +52,9 @@
   public SingleClusterMonitor(Map<String, Object> props, String name) throws Exception {
     _name = name;
     _topicManagementService = new TopicManagementService(props, name);
+    CompletableFuture topicPartitionReady = _topicManagementService.topicPartitionReady();
```
sutambe: `CompletableFuture` or `CompletableFuture<Void>`?

src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -105,25 +104,26 @@ public ConsumeService(Map<String, Object> props, String name) throws Exception {
...
-
-    MetricConfig metricConfig = new MetricConfig().samples(60).timeWindow(1000, TimeUnit.MILLISECONDS);
-    List<MetricsReporter> reporters = new ArrayList<>();
-    reporters.add(new JmxReporter(JMX_PREFIX));
-    Metrics metrics = new Metrics(metricConfig, reporters, new SystemTime());
-    Map<String, String> tags = new HashMap<>();
-    tags.put("name", _name);
-    _adminClient = AdminClient.create(props);
-    _sensors = new ConsumeMetrics(metrics, tags, topic);
+    topicPartitionReady.thenRun(() -> {
```
sutambe: It is hard to the understand flow of future completions and its consequences. This constructor sneakily chains a runnable to `topicPartitionReady` future _inside_ the constructor. It's much better to refactor the _action_ from _chaining_. One possibility is as follows.
1. Separate out this large lambda in a method of its own.
2. Don't pass the future in constructor. 
3. When the top-level future is created in `MultiClusterTopicManagementService` (ideally not in the constructor), pass a small lambda calling the method extracted in step 1. It's **much** better to chain futures such that all chains (if possible) are visible in one small piece of code.

src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -221,13 +221,23 @@ public boolean isRunning() {
...
     private final Sensor _recordsDelayed;
 
-    ConsumeMetrics(final Metrics metrics, final Map<String, String> tags, String topicName) throws ExecutionException, InterruptedException {
-      int partitionCount = 0;
-      DescribeTopicsResult describeTopicsResult = _adminClient.describeTopics(Collections.singleton(topicName));
-      Map<String, KafkaFuture<TopicDescription>> values =  describeTopicsResult.values();
-      partitionCount = values.get(topicName).get().partitions().size();
-      Sensor topicPartitionCount = metrics.sensor("topic-partitions");
-      topicPartitionCount.add(new MetricName("topic-partitions-count", METRIC_GROUP_NAME, "The total number of partitions for the topic.", tags), new Total(partitionCount));
+    ConsumeMetrics(final Metrics metrics, final Map<String, String> tags, String topicName, CompletableFuture<Void> topicPartitionReady) {
```
sutambe: ditto here. The lambda is too long to understand its purpose. Don't call `.thenRun` in the constructor---as long as it is not a constructor! Use it in the caller of the constructor. Really.

src/main/java/com/linkedin/kmf/services/MultiClusterTopicManagementService.java
```diff
@@ -91,18 +92,19 @@ public MultiClusterTopicManagementService(Map<String, Object> props, String serv
...
-      public Thread newThread(Runnable r) {
-        return new Thread(r, _serviceName + "-multi-cluster-topic-management-service");
-      }
-    });
+    _executor = Executors.newSingleThreadScheduledExecutor(
+      r -> new Thread(r, _serviceName + "-multi-cluster-topic-management-service"));
+    _topicPartitionReady.complete(null);
   }
 
   public CompletableFuture<Void> topicManagementReady() {
```
sutambe: Perhaps this method should initiate the async operation and return the future.

src/main/java/com/linkedin/kmf/services/MultiClusterTopicManagementService.java
```diff
@@ -66,21 +65,23 @@
...
   private static final String METRIC_GROUP_NAME = "topic-management-service";
+  private final CompletableFuture<Void> _topicPartitionReady = new CompletableFuture<>();
   private final AtomicBoolean _isRunning = new AtomicBoolean(false);
   private final String _serviceName;
   private final Map<String, TopicManagementHelper> _topicManagementByCluster;
   private final int _scheduleIntervalMs;
   private final long _preferredLeaderElectionIntervalMs;
   private final ScheduledExecutorService _executor;
-  final private CompletableFuture<Void> _completableFuture;
+  private final CompletableFuture<Void> _completableFuture;
```
sutambe: Why is it necessary to store the future as a member?

src/main/java/com/linkedin/kmf/services/TopicManagementService.java
```diff
@@ -23,10 +23,16 @@
  */
 public class TopicManagementService implements Service {
   private final MultiClusterTopicManagementService _multiClusterTopicManagementService;
+  CompletableFuture<Void> _topicPartitionReady = new CompletableFuture<>();
 
   public TopicManagementService(Map<String, Object> props, String serviceName) throws Exception {
     Map<String, Object> serviceProps = createMultiClusterTopicManagementServiceProps(props, serviceName);
     _multiClusterTopicManagementService = new MultiClusterTopicManagementService(serviceProps, serviceName);
+    _topicPartitionReady.complete(null);
```
sutambe: Very, _very_ sneaky. Please refactor.

src/main/java/com/linkedin/kmf/services/MultiClusterTopicManagementService.java
andrewchoi5: Will put in the backlog the Asynchronous operation for `topicmanagementReady()`.

src/main/java/com/linkedin/kmf/services/ConsumeService.java
andrewchoi5: That's correct. I was debating why the original code has a constructor of huge chunk that contains a large lambda. In the separate PR, we will factor out this large lambda.

src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
andrewchoi5: Amended.

src/main/java/com/linkedin/kmf/apps/MultiClusterMonitor.java
andrewchoi5: The new AdminClient, which is eventually consistent, renders a race condition for consumeService, wherein the produce availability indicates 1 but the consumer availability shows 0 availability without eventually converging. Thus, the topic partitions are ready in time for produceService but not for consumeService.

src/main/java/com/linkedin/kmf/tests/BasicEndToEndTest.java
```diff
@@ -44,8 +44,9 @@
   public BasicEndToEndTest(Map<String, Object> props, String name) throws Exception {
     _name = name;
     _topicManagementService = new TopicManagementService(props, name);
+    CompletableFuture<Void> topicPartitionReady = _topicManagementService.topicPartitionReady();
     _produceService = new ProduceService(props, name);
-    _consumeService = new ConsumeService(props, name);
+    _consumeService = new ConsumeService(props, name, topicPartitionReady);
```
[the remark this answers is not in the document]
andrewchoi5: As discussed in person, resolving this conversation.

src/main/java/com/linkedin/kmf/apps/MultiClusterMonitor.java
andrewchoi5: Sure. Lambda refactoring is already in the backlog, since the constructors for producers and consumers are too large in general as well.

src/main/java/com/linkedin/kmf/apps/MultiClusterMonitor.java
andrewchoi5: Asynchronous operations will be refactored out of constructor in a separate PR.

src/main/java/com/linkedin/kmf/services/TopicManagementService.java
andrewchoi5: As described above, the asynchronous operations refactoring will be included in a PR

src/main/java/com/linkedin/kmf/apps/MultiClusterMonitor.java
andrewchoi5: AdminClient topic creation is relatively slow and and thus is a cause for the race condition problems.

src/main/java/com/linkedin/kmf/services/MultiClusterTopicManagementService.java
andrewchoi5:
```
 public CompletableFuture<Void> topicManagementReady() {
    return _completableFuture;
  }```

The `Topicmanagementservice` and `multiclustermonitor` makes a call to the topicManagementReady() method in order to retrieve the completableFuture.

src/main/java/com/linkedin/kmf/services/ConsumeService.java
andrewchoi5: Discussed in a previous conversation.

src/main/java/com/linkedin/kmf/apps/MultiClusterMonitor.java
andrewchoi5: Stylistic concerns including method and variable naming conventions/lambda will be included in a separate PR.

PR #183, opened 2019-12-12
JsonFromGenericRecord Method in Utils class Enable public access

andrewchoi5:
** jsonFromGenericRecord(..)** in `Utils` class enable `public` access such that external `dependencies` can have access to this method when needed.

### DIFF
--- a/src/main/java/com/linkedin/kmf/common/Utils.java
+++ b/src/main/java/com/linkedin/kmf/common/Utils.java
@@ -150,7 +150,7 @@ public class Utils {
     return record;
   }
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/linkedin/kmf/common/Utils.java
```diff
@@ -150,7 +150,7 @@ public static GenericRecord genericRecordFromJson(String message) {
     return record;
   }
```
andrewchoi5: @smccauliff merging this PR as this is a small change that unblocks internal MP testing.

Issue #178, opened 2019-12-09
Kafka SSL handhake failed


KhoiDinh:
I am trying to connect kafka monitor to my sasl_ssl configured kafka cluster. My kafka cluster currently works with sasl_ssl but kafka monitor keeps on failed. This is what I get when I try to run kafka monitor:

>  [2019-12-09 15:19:10,887] INFO MultiClusterTopicManagementServiceConfig values: 
> 	topic = my-topic
> 	topic-management.preferred.leader.election.check.interval.ms = 300000
> 	topic-management.rebalance.interval.ms = 600000
>  (com.linkedin.kmf.services.configs.MultiClusterTopicManagementServiceConfig)
> [2019-12-09 15:19:10,941] INFO TopicManagementServiceConfig values: 
> 	topic = my-topic
> 	topic-management.minPartitionNum = 1
> 	topic-management.partitionsToBrokersRatio = 2.0
> 	topic-management.replicationFactor = 1
> 	topic-management.topicCreationEnabled = true
> 	topic-management.topicFactory.class.name = com.linkedin.kmf.topicfactory.DefaultTopicFactory
> 	zookeeper.connect = 10.15.164.233:2181
>  (com.linkedin.kmf.services.configs.TopicManagementServiceConfig)
> [2019-12-09 15:19:11,005] INFO ProduceServiceConfig values: 
> 	bootstrap.servers = 10.15.164.233:9093
> 	produce.latency.percentile.granularity.ms = 1
> 	produce.latency.percentile.max.ms = 5000
> 	produce.partitioner.class = class com.linkedin.kmf.partitioner.NewKMPartitioner
> 	produce.producer.class = com.linkedin.kmf.producer.NewProducer
> 	produce.producer.id = kmf-producer
> 	produce.record.delay.ms = 100
> 	produce.record.size.byte = 100
> 	produce.sync = true
> 	produce.thread.num = 5
> 	produce.treat.zero.throughput.as.unavailable = true
> 	topic = my-topic
> 	zookeeper.connect = 10.15.164.233:2181
>  (com.linkedin.kmf.services.configs.ProduceServiceConfig)
> [2019-12-09 15:19:11,146] INFO single-cluster-monitor/ProduceService is initialized. (com.linkedin.kmf.services.ProduceService)
> [2019-12-09 15:19:11,152] INFO ConsumeServiceConfig values: 
> 	bootstrap.servers = 10.15.164.233:9093
[...]

PR #187, opened 2019-12-18
[WIP] Enable SSL Configurations for ISSUE #178

andrewchoi5:
This will be merged **after** the PR `(https://github.com/linkedin/kafka-monitor/pull/181)` is merged. and then this _PR_ rebased.

Addresses the issue:
> https://github.com/linkedin/kafka-monitor/issues/178


src/main/java/com/linkedin/kmf/common/Utils.java
```diff
@@ -45,6 +50,42 @@
...
+    props.put(SslConfigs.SSL_KEYSTORE_PASSWORD_CONFIG, adminClientConfig.getString(SslConfigs.SSL_KEYSTORE_PASSWORD_CONFIG));
+    props.put(SslConfigs.SSL_KEY_PASSWORD_CONFIG, adminClientConfig.getString(SslConfigs.SSL_KEY_PASSWORD_CONFIG));
+    props.put(SslConfigs.SSL_TRUSTSTORE_LOCATION_CONFIG, adminClientConfig.getString(SslConfigs.SSL_TRUSTSTORE_LOCATION_CONFIG));
+    props.put(SslConfigs.SSL_TRUSTSTORE_PASSWORD_CONFIG, adminClientConfig.getString(SslConfigs.SSL_TRUSTSTORE_PASSWORD_CONFIG));
+    props.put(SslConfigs.SSL_KEYSTORE_TYPE_CONFIG, adminClientConfig.getString(SslConfigs.SSL_KEYSTORE_TYPE_CONFIG));
+    props.put(SslConfigs.SSL_SECURE_RANDOM_IMPLEMENTATION_CONFIG, adminClientConfig.getString(SslConfigs.SSL_SECURE_RANDOM_IMPLEMENTATION_CONFIG));
+    props.put(SslConfigs.SSL_TRUSTMANAGER_ALGORITHM_CONFIG, adminClientConfig.getString(SslConfigs.SSL_TRUSTMANAGER_ALGORITHM_CONFIG));
+    props.put(SslConfigs.SSL_TRUSTSTORE_TYPE_CONFIG, adminClientConfig.getString(SslConfigs.SSL_TRUSTSTORE_TYPE_CONFIG));
+    return props;
+  }
```
smccauliff: What does this do that can not be done with the existing configuration system?

This does not address the user issue as they are using SASL.

When the Kafka producers and consumers add new properties this block of code will need to be changed which means the code is more brittle when upgrading between versions of Kafka.

src/main/java/com/linkedin/kmf/apps/MultiClusterMonitor.java
```diff
@@ -70,28 +72,31 @@ public MultiClusterMonitor(Map<String, Object> props, String name) throws Except
...
   }
 
   @Override
   public boolean isRunning() {
-    return _topicManagementService.isRunning() && _produceService.isRunning() && _consumeService.isRunning();
+    return _multiClusterTopicManagementService.isRunning() && _produceService.isRunning() && _consumeService.isRunning();
   }
 
   @Override
   public void awaitShutdown() {
```
Lincong: I know that it is out of the scope of this PR. But I think changing the `awaitShutdown()` method in the `App` interface and make it take some timeout parameter might be a better practice. For example, refer to the `ExecutorService` API `boolean awaitTermination(long timeout, TimeUnit unit)`. No need to change now. But it would be great if you can consider this change to be on the backlog.

src/main/java/com/linkedin/kmf/services/MultiClusterTopicManagementService.java
```diff
@@ -262,7 +262,9 @@ public void run() {
       Map topicFactoryConfig = props.containsKey(TopicManagementServiceConfig.TOPIC_FACTORY_PROPS_CONFIG) ?
           (Map) props.get(TopicManagementServiceConfig.TOPIC_FACTORY_PROPS_CONFIG) : new HashMap();
       _topicFactory = (TopicFactory) Class.forName(topicFactoryClassName).getConstructor(Map.class).newInstance(topicFactoryConfig);
+
       _adminClient = constructAdminClient(props);
+      LOG.info("{} configs: {}", _adminClient.getClass().getSimpleName(), _adminClient);
```
Lincong: The second argument to the logger is of type `AdminClient`. Do you mean to use `props` instead of `_adminClient `?

src/main/java/com/linkedin/kmf/services/MultiClusterTopicManagementService.java
```diff
@@ -66,21 +66,23 @@
...
   private static final String METRIC_GROUP_NAME = "topic-management-service";
+  private final CompletableFuture<Void> _topicPartitionReady = new CompletableFuture<>();
   private final AtomicBoolean _isRunning = new AtomicBoolean(false);
   private final String _serviceName;
   private final Map<String, TopicManagementHelper> _topicManagementByCluster;
   private final int _scheduleIntervalMs;
   private final long _preferredLeaderElectionIntervalMs;
   private final ScheduledExecutorService _executor;
-  final private CompletableFuture<Void> _completableFuture;
+  private final CompletableFuture<Void> _completableFuture;
```
Lincong: Again, can we not name a variable by its type (unless there is a very good reason for doing it)?

src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
```diff
@@ -77,6 +79,15 @@ public void stop() {
 
   @Override
   public boolean isRunning() {
```
Lincong: Nit: maybe consider the following code
```
    boolean isRunning = true;

    if (!_topicManagementService.isRunning()) {
      isRunning = false;
      LOG.info("_topicManagementService not running.");
    }
    if (!_produceService.isRunning()) {
      isRunning = false;
      LOG.info("_produceService not running.");
    }
    if (!_consumeService.isRunning()) {
      isRunning = false;
      LOG.info("_consumeService not Running.");
    }
    return isRunning;
```

src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -57,16 +58,16 @@
       ConsumeServiceConfig.ZOOKEEPER_CONNECT_CONFIG};
 
   private final String _name;
-  private final ConsumeMetrics _sensors;
+  private ConsumeMetrics _sensors;
   private final KMBaseConsumer _consumer;
-  private final Thread _thread;
+  private Thread _thread;
```
Lincong: Maybe give this thread a more informative name?

src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
andrewchoi5: amended.

src/main/java/com/linkedin/kmf/apps/MultiClusterMonitor.java
andrewchoi5: This could be in my backlog collections for revisions in the future.

Issue #195, opened 2020-01-09
Broken ConsumeService on latest release


d1egoaz:
According to this: https://github.com/linkedin/kafka-monitor/issues/139
and this: https://github.com/linkedin/kafka-monitor/wiki/App-Configuration

We could have something like this configured:
```json
"consume-service": {
    "class.name": "com.linkedin.kmf.services.ConsumeService",
    "topic": "kafka-monitor-topic",
    "zookeeper.connect": "localhost:2181",
    "bootstrap.servers": "localhost:9092",
    "consume.latency.sla.ms": "20000",
    "consume.consumer.props": {
    }
```

However, this commit change the method's arguments on `ConsumeService`
https://github.com/linkedin/kafka-monitor/commit/f1a21c9a16df1d274f424111ae0fab69e6f85d0b
![image](https://user-images.githubusercontent.com/950087/72084896-58a5c000-32d2-11ea-9575-eb41d86040eb.png)

and now it's failing with:
```
Exception in thread "main" java.lang.NoSuchMethodException: com.linkedin.kmf.services.ConsumeService.<init>(java.util.Map, java.lang.String)
        at java.lang.Class.getConstructor0(Class.java:3082)
        at java.lang.Class.getConstructor(Class.java:1825)
        at com.linkedin.kmf.KafkaMonitor.<init>(KafkaMonitor.java:71)
        at com.linkedin.kmf.KafkaMonitor.main(KafkaMonitor.java:166)
```
as the executor is only sending 2 arguments, instead of the 3 the method is now expecting.

andrewchoi5:
there is a PR that's ready to be released regarding the constructor parameters for new KMF. thanks.

andrewchoi5:
Hello,

Could you please include the full stacktrace as well as the command you've actually executed to produce this?

andrewchoi5:
A Pull Request will be ready to resolve this NoSuchMethodException issue for ConsumeService. Thanks.

andrewchoi5:
`https://github.com/linkedin/kafka-monitor/pull/196`

d1egoaz:
Great 👍 thank you !

andrewchoi5:
No problem! Thank you 👍

PR #196, opened 2020-01-16
Commit Availability Service

andrewchoi5:
**Commit Availability Service.**
1 - Addresses: Broken ConsumeService on latest release #195 https://github.com/linkedin/kafka-monitor/issues/195
2 - Addresses: https://github.com/linkedin/kafka-monitor/issues/204 
3 - Performs Consumer Offset Commit Availability using Metric Sensors. Does periodically, every few x seconds.
4 - In doing this, public interface `ConsumerFactory` is created to build consumer and pass into `consumeService`.

WIP as a follow up PR : Commit `Latency` Service


src/main/java/com/linkedin/kmf/KafkaMonitor.java
```diff
@@ -79,8 +90,11 @@ public KafkaMonitor(Map<String, Map> testProps) throws Exception {
     List<MetricsReporter> reporters = new ArrayList<>();
     reporters.add(new JmxReporter(JMX_PREFIX));
     Metrics metrics = new Metrics(new MetricConfig(), reporters, new SystemTime());
-    metrics.addMetric(metrics.metricName("offline-runnable-count", METRIC_GROUP_NAME, "The number of Service/App that are not fully running"),
-      (config, now) -> _offlineRunnables.size());
+    Map<String, String> tags = new HashMap<>();
```
Lincong: Nit: If there is only one value in the map, you can use `Map<String, String> tags = new HashMap<>(1)`

src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
```diff
@@ -61,6 +64,7 @@ public SingleClusterMonitor(Map<String, Object> props, String name) throws Excep
   @Override
   public void start() {
     _topicManagementService.start();
+    _commitAvailabilityService.start();
     CompletableFuture<Void> completableFuture = _topicManagementService.topicManagementResult();
     completableFuture.thenRun(() -> {
```
Lincong: I know this line is not introduced as part of this PR. But I am just wondering why making the start of the produceService and consumerService async? Is the goal here to make the method `start()` return faster? I think so services their `start()` methods called synchronously whereas others don't. It feels a little strange.

Another problem with having async start is that what if `_produceService.start();` or `_consumeService.start();` throws exception(s)? No code is going to handle it and the whole `SingleClusterMonitor` seems to have started. But actually not (inconsistent state).

src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
```diff
@@ -85,6 +90,10 @@ public boolean isRunning() {
       isRunning = false;
       LOG.info("_topicManagementService not running.");
     }
+    if (!_commitAvailabilityService.isRunning()) {
```
Lincong: Why not return false when we encounter the first not-running service?

src/main/java/com/linkedin/kmf/KafkaMonitor.java
andrewchoi5: Updated.

src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
andrewchoi5: Please reference the previous Pull Requests for justifications to asynchronous calls. They have to do with the fact that  `_topicManagementResult` has to `complete()` within `MultiClusterTopicManagementService` class before the `produceService` and `consumeService` can start otherwise the ordering of these calls get disoriented causing failures.;

src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
andrewchoi5: This codestyle decision is actually a product of your suggestion back in PR number 187 (https://github.com/linkedin/kafka-monitor/pull/187)

Originally, the `isRunning` method returned false when the first service was not, in fact, running.

image

src/main/java/com/linkedin/kmf/services/CommitAvailabilityService.java
```diff
@@ -0,0 +1,238 @@
...
+    }
+  }
+
+  private OffsetAndMetadata commitAndRetrieveOffsets(TopicPartition topicPartition, Map<TopicPartition, OffsetAndMetadata> offsetMap) throws Exception {
+    final AtomicBoolean callbackFired = new AtomicBoolean(false);
+    final AtomicReference<Exception> offsetCommitIssue = new AtomicReference<>(null);
+    OffsetAndMetadata committed = null;
+    long now = System.currentTimeMillis();
+    long deadline = now + TimeUnit.MINUTES.toMillis(1);
+    while (System.currentTimeMillis() < deadline) {
```
smccauliff: Why have a deadline?  Why not just have a thread that periodically commits offsets and then use the callback to measure commit time?

config/kafka-monitor.properties
```diff
@@ -43,7 +43,7 @@
 {
   "single-cluster-monitor": {
     "class.name": "com.linkedin.kmf.apps.SingleClusterMonitor",
-    "topic": "kafka-monitor-topic",
+    "topic": "kafka-monitor-topic-name",
```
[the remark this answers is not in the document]
andrewchoi5: reverted. was testing with different topic name locally.

src/main/java/com/linkedin/kmf/KafkaMonitor.java
```diff
@@ -40,7 +42,7 @@
   private static final Logger LOG = LoggerFactory.getLogger(KafkaMonitor.class);
   public static final String CLASS_NAME_CONFIG = "class.name";
   private static final String METRIC_GROUP_NAME = "kafka-monitor";
-  private static final String JMX_PREFIX = "kmf";
+  private static final String JMX_PREFIX = "kmf.services";
```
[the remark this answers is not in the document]
andrewchoi5: this change was for the purposes of alignment of with other sensors metrics. will amend.

src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
```diff
@@ -99,6 +115,7 @@ public boolean isRunning() {
   @Override
   public void awaitShutdown() {
     _topicManagementService.awaitShutdown();
+    _commitAvailabilityService.awaitShutdown();
     _produceService.awaitShutdown();
     _consumeService.awaitShutdown();
   }
```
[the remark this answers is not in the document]
andrewchoi5: Done

src/main/java/com/linkedin/kmf/KafkaMonitor.java
```diff
@@ -79,8 +90,11 @@ public KafkaMonitor(Map<String, Map> testProps) throws Exception {
     List<MetricsReporter> reporters = new ArrayList<>();
     reporters.add(new JmxReporter(JMX_PREFIX));
     Metrics metrics = new Metrics(new MetricConfig(), reporters, new SystemTime());
-    metrics.addMetric(metrics.metricName("offline-runnable-count", METRIC_GROUP_NAME, "The number of Service/App that are not fully running"),
-      (config, now) -> _offlineRunnables.size());
+    Map<String, String> tags = new HashMap<>(1);
+    tags.put("name", "kafka-monitor");
```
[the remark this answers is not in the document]
andrewchoi5: All other metrics and sensors have the following structure,
```kmf.services:name=single-cluster-monitor,type=produce-service:produce-availability-avg=1.0
kmf.services:name=single-cluster-monitor,type=consume-service:consume-availability-avg=1.0
kmf.services:name=single-cluster-monitor,type=commit-availability-service:offsets-committed-avg=1.0
kmf.services:name=single-cluster-monitor,type=commit-availability-service:failed-commit-offsets-avg=0.0
kmf.services:name=single-cluster-monitor,type=produce-service:records-produced-total=148.0
kmf.services:name=single-cluster-monitor,type=consume-service:records-consumed-total=116.0
kmf.services:name=single-cluster-monitor,type=consume-service:records-lost-total=0.0```

except for offline runnables:
kmf:name=kafka-monitor,type=kafka-monitor:offline-runnable-count=0.0

This fixes such inconsistency.

File: src/main/java/com/linkedin/kmf/services/CommitAvailabilityService.java (hunk 4 above)
Comment (andrewchoi5) [reply src/main/java/com/linkedin/kmf/services/CommitAvailabilityService.java]: Revised to utilize period offset commits

File: src/main/java/com/linkedin/kmf/services/CommitAvailabilityService.java [hunk 9]
```diff
@@ -0,0 +1,238 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testng.Assert;
+
+import static org.testng.Assert.assertNotNull;
+
+/**
+ * CommitAvailabilityService measures the availability of consume offset commits to the Kafka broker.
+ */
+public class CommitAvailabilityService implements Service {
```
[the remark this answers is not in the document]
andrewchoi5: Would we always want to run Commit Availability Service when KMF is running Consume Service? The expectations are that commit availabilities are needed on some occasions.

src/main/java/com/linkedin/kmf/services/CommitAvailabilityService.java
```diff
@@ -175,11 +175,9 @@ private void initiateCommit() throws Exception {
   private OffsetAndMetadata commitAndRetrieveOffsets(TopicPartition topicPartition, Map<TopicPartition, OffsetAndMetadata> offsetMap) throws Exception {
     final AtomicBoolean callbackFired = new AtomicBoolean(false);
     final AtomicReference<Exception> offsetCommitIssue = new AtomicReference<>(null);
-    OffsetAndMetadata committed = null;
-    long now = System.currentTimeMillis();
-    long deadline = now + TimeUnit.MINUTES.toMillis(1);
-    while (System.currentTimeMillis() < deadline) {
-      // Call commitAsync, wait for a NON-NULL return value (see https://issues.apache.org/jira/browse/KAFKA-6183)
+    OffsetAndMetadata committed;
+    while (true) {
```
smccauliff: Should this check if the service has stopped?

src/main/java/com/linkedin/kmf/services/CommitAvailabilityService.java
```diff
@@ -0,0 +1,238 @@
...
+        commitAndRetrieveOffsets(topicPartition, _offsetsToCommit);
+
+      } catch (KafkaException kafkaException) {
+        LOG.error("Exception while trying to to async commit.", kafkaException);
+        _commitAvailabilityMetrics._failedCommitOffsets.record();
+      }
+    }
+  }
+
+  private OffsetAndMetadata commitAndRetrieveOffsets(TopicPartition topicPartition, Map<TopicPartition, OffsetAndMetadata> offsetMap) throws Exception {
```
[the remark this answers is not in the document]
andrewchoi5: Since it’s impossible to add testing for commitRetrieveOffsets without initializing a new CommitAvailabilityService instance in the Test  class. Majority of KMF code is badly structured (when originally written since its inception) for unit testing.

src/main/java/com/linkedin/kmf/services/CommitAvailabilityMetrics.java
```diff
@@ -0,0 +1,54 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(CommitAvailabilityMetrics.class);
+
+  /**
+   * Metrics for Calculating the offset commit availability of a consumer.
+   * @param metrics the commit offset metrics
+   * @param tags the tags associated, i.e) kmf.services:name=single-cluster-monitor
+   */
+  CommitAvailabilityMetrics(final Metrics metrics, final Map<String, String> tags) {
+    LOG.info("{} called.", this.getClass().getSimpleName());
+    _offsetsCommitted = metrics.sensor("offsets-committed");
```
smccauliff: I'm not sure how quantile metrics are computed with the metric library being used, but perhaps there should be  some kind of 99th percentile commit time.  Is this the CommitLatencyService?

src/main/java/com/linkedin/kmf/services/ConsumeMetrics.java
```diff
@@ -0,0 +1,120 @@
...
+      DescribeTopicsResult describeTopicsResult = adminClient.describeTopics(Collections.singleton(topicName));
+      Map<String, KafkaFuture<TopicDescription>> topicResultValues = describeTopicsResult.values();
+      KafkaFuture<TopicDescription> topicDescriptionKafkaFuture = topicResultValues.get(topicName);
+      TopicDescription topicDescription = null;
+      try {
+        topicDescription = topicDescriptionKafkaFuture.get();
+      } catch (InterruptedException | ExecutionException e) {
+        LOG.error("Exception occurred while retrieving the topic description.", e);
+      }
+      assert topicDescription != null;
```
smccauliff: Asserts are special instructions which are not normally executed by the JVM unless the -ea flag is set when the JVM starts up.  The default is is that the flag is not set.  So asserts in Java are really only useful if the check the assert is guarding is  something expensive like an O(n) check.  For O(1) checks just throwing an exception is almost always better.

src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -149,15 +147,52 @@ private void consume() throws Exception {
...
+        };
+
+        if (_offsetsToCommit != null) {
+          _consumer.commitAsync(_offsetsToCommit, commitCallback);
+        } else {
+          _consumer.commitAsync(commitCallback);
+        }
+        _commitAvailabilityMetrics._offsetsCommitted.record();
+
+        org.testng.Assert.assertNull(offsetCommitIssues.get(), "Offset commit failed.");
```
smccauliff: testng should not be used within actual production code.

If having a method that checks for a condition is more clear then perhaps the google guava library is the way to go.  https://github.com/google/guava/wiki/PreconditionsExplained  It's probably ok to add a dependency on that library KMF does not already have this dependency.

src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -110,7 +106,8 @@ public ConsumeService(Map<String, Object> props, String name, CompletableFuture<
     if (props.containsKey(ConsumeServiceConfig.CONSUMER_PROPS_CONFIG)) {
       props.forEach(consumerProps::putIfAbsent);
     }
-    _consumer = (KMBaseConsumer) Class.forName(consumerClassName).getConstructor(String.class, Properties.class).newInstance(topic, consumerProps);
+    _consumer = (KMBaseConsumer) Class.forName(consumerClassName).getConstructor(String.class, Properties.class).newInstance(_topic, consumerProps);
```
smccauliff: Because the consume service is creating it's own consumer rather than using an injected consumer it's not possible to mock the consumer without lots of effort.  This is one of the challenges with unit testing this class.  If the consume service either takes a consumer or  consumer factory as a parameter then mocking becomes possible, which means unit testing becomes possible.

andrewchoi5: In order to render unit testing possible for `ConsumeService`, I will make the `consumeService` take in consumer / consumer factory as an argument in a follow PR if that's feasible and okay with you.

src/main/java/com/linkedin/kmf/services/ConsumeMetrics.java
andrewchoi5: Agreed. Revised.

src/main/java/com/linkedin/kmf/services/CommitAvailabilityMetrics.java
andrewchoi5: The class in discussion is for the `CommitAvailabilityMetrics`. The `CommitLatencyService` will be sampling and determining the 99th percentile commit time for its calculation.

src/main/java/com/linkedin/kmf/services/ConsumeService.java
andrewchoi5: Amended to not use testng library within the prod codebase.

src/test/java/com/linkedin/kmf/services/ConsumeServiceTest.java
```diff
@@ -0,0 +1,108 @@
...
+ * Unit Testing for the Consume Service Class.
+ */
+public class ConsumeServiceTest {
+  private static final String BROKER_LIST = "localhost:9092";
+  private static final String ZK_CONNECT = "localhost:2181";
+  private static final String TOPIC = "kafka-monitor-topic-test";
+
+  @Test
+  public void lifecycleTest()
+      throws ClassNotFoundException, NoSuchMethodException, InstantiationException, IllegalAccessException,
```
smccauliff: Giant exception lists are hard to read and not very useful.  Also the test framework treats all exceptions the same unless it is told not to.  So it's safe to make this just throw Exception.

src/test/java/com/linkedin/kmf/services/ConsumeServiceTest.java
```diff
@@ -0,0 +1,108 @@
...
+             InvocationTargetException {
+
+    ConsumeService consumeService = consumeService();
+
+    /* Nothing should be started */
+    org.testng.Assert.assertFalse(consumeService.isRunning());
```
smccauliff: No one writes test code like org.testng.Assert; it's either Assert.assertFalse or do the static import of the methods used for assertions.

src/main/java/com/linkedin/kmf/services/ConsumerFactory.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.Map;
+import java.util.Properties;
+import java.util.Random;
+import org.apache.kafka.clients.admin.AdminClient;
+import org.apache.kafka.clients.consumer.ConsumerConfig;
+import org.apache.kafka.common.config.ConfigException;
+import org.apache.kafka.common.serialization.StringDeserializer;
+
+
+public class ConsumerFactory {
```
smccauliff: If ConsumerFactory is an interface then it will be a little easier to mock.  This implementation then moves to ConsumerFactoryImpl (or whatever you want to call it).

src/test/java/com/linkedin/kmf/services/ConsumeServiceTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+    /* Should be allowed to shutdown more than once. */
+    consumeService.awaitShutdown();
+    consumeService.awaitShutdown();
+    org.testng.Assert.assertFalse(consumeService.isRunning());
+
+  }
+
+
+  private ConsumeService consumeService()
```
smccauliff: Once the ConsumerFactory is an interface most of this code can go away and be replaced with code that returns a mocked consumer.

src/test/java/com/linkedin/kmf/services/ConsumeServiceTest.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicReference;
+import org.apache.kafka.clients.consumer.ConsumerConfig;
+import org.testng.annotations.Test;
+
+
+/**
+ * Unit Testing for the Consume Service Class.
+ */
+public class ConsumeServiceTest {
```
smccauliff: Thanks for creating this test.  While this looks like it will test lifecycle will it test any of the commit availability code?

src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -149,15 +98,50 @@ private void consume() throws Exception {
...
+        final AtomicReference<Exception> offsetCommitIssues = new AtomicReference<>(null);
+        OffsetAndMetadata offsetAndMetadata;
+        /* Call commitAsync, wait for a NON-NULL return value (see https://issues.apache.org/jira/browse/KAFKA-6183) */
+        OffsetCommitCallback commitCallback = new OffsetCommitCallback() {
+          @Override
+          public void onComplete(Map<TopicPartition, OffsetAndMetadata> topicPartitionOffsetAndMetadataMap, Exception exception) {
+            if (exception != null) {
+              offsetCommitIssues.set(exception);
+            }
+            callbackFired.set(true);
```
smccauliff: When does this value get used?

src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -149,15 +98,50 @@ private void consume() throws Exception {
...
+          }
+        };
+
+        if (_offsetsToCommit != null && !_offsetsToCommit.isEmpty()) {
```
smccauliff: Maybe don't commit offsets every message as it's kind of an extreme use case.  Perhaps select a time period, like every second or 5 seconds for  committing an offset.

src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -149,15 +98,50 @@ private void consume() throws Exception {
...
+          _baseConsumer.commitAsync(_offsetsToCommit, commitCallback);
+        } else {
+          _baseConsumer.commitAsync(commitCallback);
+        }
+        _commitAvailabilityMetrics._offsetsCommitted.record();
```
smccauliff: Shouldn't this be in the offset commit Callback?

src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -149,15 +98,50 @@ private void consume() throws Exception {
...
+
+        offsetAndMetadata = _baseConsumer.committed(topicPartition);
+        if (offsetAndMetadata != null) {
+          _offsetsToCommit.putIfAbsent(topicPartition, offsetAndMetadata);
+        }
```
smccauliff: Won't this commit the same offsets every time?

src/main/java/com/linkedin/kmf/KafkaMonitor.java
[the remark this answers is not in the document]
andrewchoi5: Revert.

src/main/java/com/linkedin/kmf/KafkaMonitor.java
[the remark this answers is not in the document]
andrewchoi5: sgtm.

src/test/java/com/linkedin/kmf/services/ConsumeServiceTest.java
andrewchoi5: Revised.

build.gradle
```diff
@@ -41,8 +41,10 @@ allprojects {
     compile 'net.savantly:graphite-client:1.1.0-RELEASE'
     compile 'com.timgroup:java-statsd-client:3.0.1'
     compile 'com.signalfx.public:signalfx-codahale:0.0.47'
+    compile 'org.mockito:mockito-core:2.24.0'
```
smccauliff: Why does this need a compile time dependency on mockito?  testCompile should be sufficient.

src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
```diff
@@ -245,7 +254,7 @@ private static ArgumentParser argParser() {
   public static void main(String[] args) throws Exception {
     ArgumentParser parser = argParser();
     if (args.length == 0) {
-      System.out.println(parser.formatHelp());
+      LOG.info(parser.formatHelp());
```
smccauliff: This is command line help and needs to be reported to the user on standard output or standard error.

src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -7,157 +7,148 @@
...
+          }
+
+          /* Record the current time for the committed consumer offset */
+          _baseConsumer.updateLastCommit();
+        }
+
+        offsetAndMetadata = _baseConsumer.committed(topicPartition);
+        if (offsetAndMetadata != null) {
+          _offsetsToCommit.put(topicPartition, offsetAndMetadata);
+        }
```
smccauliff: I'm still confused about why this is needed and how the offsets for the consumer will not get stuck since it would always be committing the value it was committing before.

src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -7,157 +7,148 @@
...
+
+      /* Commit availability and commit latency service */
+      try {
+        TopicPartition topicPartition = new TopicPartition(_topic, partition);
+        final AtomicReference<Exception> offsetCommitIssues = new AtomicReference<>(null);
+        OffsetAndMetadata offsetAndMetadata;
+        /* Call commitAsync, wait for a NON-NULL return value (see https://issues.apache.org/jira/browse/KAFKA-6183) */
+        OffsetCommitCallback commitCallback = (topicPartitionOffsetAndMetadataMap, exception) -> {
+          if (exception != null) {
+            offsetCommitIssues.set(exception);
```
smccauliff: Is offsetCommitIssues ever used?  Should the following be called?

```java
        LOG.error("Exception while trying to perform an asynchronous commit.", kafkaException);
        _commitAvailabilityMetrics._failedCommitOffsets.record();
```

src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -7,157 +7,148 @@
...
+          }
+          _commitAvailabilityMetrics._offsetsCommitted.record();
+        };
+
+        long currTimeMillis = System.currentTimeMillis();
+
+        /* 5 seconds consumer offset commit interval. */
+        long timeDiffMillis = 5000;
```
smccauliff: The more readable way of having time constants in the code is to do something like this:

```java
long timeDiffMillis = TimeUnit.SECONDS.toMillis(5);
```
@radai-rosenblatt  says there is another way using the Duration class that is more modern.

andrewchoi5: Amended.

src/main/java/com/linkedin/kmf/services/ConsumeService.java
andrewchoi5: Revised to record failedCommitOffsets

src/main/java/com/linkedin/kmf/services/ConsumeService.java
andrewchoi5: Correct. Updated.

src/test/java/com/linkedin/kmf/services/ConsumeServiceTest.java
```diff
@@ -0,0 +1,203 @@
...
+
+    ScheduledExecutorService executorService = Executors.newSingleThreadScheduledExecutor();
+    ScheduledFuture<?> scheduledFuture = executorService.schedule(new Runnable() {
+      @Override
+      public void run() {
+        Assert.assertNotNull(metrics.metrics().get(metrics.metricName("offsets-committed-total", METRIC_GROUP_NAME, tags)).metricValue());
+        Assert.assertNotEquals(metrics.metrics().get(metrics.metricName("offsets-committed-total", METRIC_GROUP_NAME, tags)).metricValue(), 0.0);
+      }
+    }, 4, TimeUnit.SECONDS);
+    Assert.assertNotNull(scheduledFuture);
```
smccauliff: This code is unlikely to be executed during the duration of the test since nothing waits for it to complete or checks that the exceptions that would be thrown from assertion failures are not there.  Was the intention something like this:
```java
   scheduledFuture.get();  // This will throw ExecutionException
```

src/test/java/com/linkedin/kmf/services/ConsumeServiceTest.java
```diff
@@ -0,0 +1,203 @@
...
+
+  }
+
+  static final class FakeConsumerFactory implements ConsumerFactory {
+    final Map<String, Object> _props;
+
+    /* Required */
+    public FakeConsumerFactory(Map<String, Object> props) {
+      _props = props;
+    }
```
smccauliff: Why is this required?  If this is not required then this class can go away as it can just be mocked.

src/test/java/com/linkedin/kmf/services/ConsumeServiceTest.java
```diff
@@ -0,0 +1,203 @@
...
+
+    @Override
+    public AdminClient adminClient() {
+      return AdminClient.create(_props);
+    }
```
smccauliff: This can be mocked

src/test/java/com/linkedin/kmf/services/ConsumeServiceTest.java
```diff
@@ -0,0 +1,203 @@
...
+
+    @Override
+    public int latencySlaMs() {
+      return 20000;
+    }
+
+    @Override
+    public KMBaseConsumer baseConsumer() {
+      return Mockito.mock(KMBaseConsumer.class);
```
smccauliff: How are expectations being added to this mocked consumer?  Shouldn't it return some messages or something.
This might be a little dated but it seems fairly comprehensive.
https://www.vogella.com/tutorials/Mockito/article.html

src/test/java/com/linkedin/kmf/services/ConsumeServiceTest.java
andrewchoi5: modified with `scheduledFuture.get();`

src/test/java/com/linkedin/kmf/services/ConsumeServiceTest.java
andrewchoi5: _Updated._

src/test/java/com/linkedin/kmf/services/ConsumeServiceTest.java
andrewchoi5: Done

src/test/java/com/linkedin/kmf/services/ConsumeServiceTest.java
```diff
@@ -0,0 +1,186 @@
...
+      public String answer(InvocationOnMock invocation) {
+        testCommittedSuccessfully = true;
+        return "Mock Offset Committed Successfully.";
+      }
+    }).when(kmBaseConsumer).commitAsync(Mockito.any());
+
+    /* avro record to KmBaseConsumer record */
+    Mockito.when(kmBaseConsumer.receive()).thenReturn(
+        new BaseConsumerRecord(TOPIC, 2, 3,
+            Utils.jsonFromFields(TOPIC, 2, 3, "producerId", 3),
```
smccauliff: The first usage of "Utils.jsonFromFields" is not needed as this parameter is just the key for the message and not the message itself (i.e. the value).  The ProduceService just sends an integer encoded as a string as the key.

src/test/java/com/linkedin/kmf/services/ConsumeServiceTest.java
```diff
@@ -0,0 +1,186 @@
...
+    consumeService.start();
+    Assert.assertTrue(consumeService.isRunning());
+
+    long threadStartDelay = 2000;
+
+    /* Thread.sleep safe to do here instead of ScheduledExecutorService
+    *  We want to sleep current thread so that consumeService can start running for enough seconds. */
+    Thread.sleep(threadStartDelay);
+    Assert.assertNotNull(metrics.metrics().get(metrics.metricName("offsets-committed-total", METRIC_GROUP_NAME, tags)).metricValue());
+    Assert.assertTrue(testCommittedSuccessfully);
```
smccauliff: This doesn't test that the ConsumeService processed the message correctly and recorded a count.

src/test/java/com/linkedin/kmf/services/ConsumeServiceTest.java
```diff
@@ -0,0 +1,186 @@
...
+    /* define return value */
+    long mockLastCommittedOffset = 0;
+    Mockito.when(kmBaseConsumer.lastCommitted()).thenReturn(mockLastCommittedOffset);
+    Mockito.doAnswer(new Answer() {
+      @Override
+      public String answer(InvocationOnMock invocation) {
+        testCommittedSuccessfully = true;
+        return "Mock Offset Committed Successfully.";
+      }
+    }).when(kmBaseConsumer).commitAsync(Mockito.any());
```
smccauliff: The callback supplied to commitAsync needs to be called inside this method.

This looks relevant.

https://www.baeldung.com/mockito-callbacks

src/test/java/com/linkedin/kmf/services/ConsumeServiceTest.java
andrewchoi5: Updated.

src/test/java/com/linkedin/kmf/services/ConsumeServiceTest.java
andrewchoi5: Done.

PR #199, opened 2020-01-16
NewConsumer deprecated polling API update and singletonList

andrewchoi5:
NewConsumer deprecated polling API update and singletonList
Signed-off-by: Andrew Choi <andchoi@linkedin.com>

### DIFF
--- a/src/main/java/com/linkedin/kmf/consumer/NewConsumer.java
+++ b/src/main/java/com/linkedin/kmf/consumer/NewConsumer.java
@@ -9,11 +9,12 @@
  */
 package com.linkedin.kmf.consumer;
 
-import org.apache.kafka.clients.consumer.ConsumerRecord;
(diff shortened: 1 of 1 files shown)


smccauliff:
It turns out that poll(Duration) and poll(long) don't have the same semantics.  Is it the intention of this patch to measure something different?

andrewchoi5:
> It turns out that poll(Duration) and poll(long) don't have the same semantics. Is it the intention of this patch to measure something different?

What do you recommend using as an alternative since the original poll() is deprecated?

For the original poll(long timeouts), since 2.0, it's been deprecated and the doc indicates using poll(Duration), which does not block beyond the timeout awaiting partition assignment. 

`https://cwiki.apache.org/confluence/display/KAFKA/KIP-266%3A+Fix+consumer+indefinite+blocking+behavior`

PR #200, opened 2020-01-17
Commenting on semantic-build-versioning.gradle file

andrewchoi5:
Commenting on semantic-build-versioning.gradle file
Signed-off-by: Andrew Choi <andchoi@linkedin.com>

### DIFF
--- a/semantic-build-versioning.gradle
+++ b/semantic-build-versioning.gradle
@@ -1,0 +1,6 @@
+
+/*
+ This file needs to be in place in order for plugin which versions jars to work correctly.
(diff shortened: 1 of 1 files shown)


smccauliff:
This file needs to be there in order for plugin which versions jars to work correctly.  Maybe just add a comment to this file saying it's needed for that plugin to function correctly.  The plugin name can be found in settings.gradle file.

smccauliff:
Perhaps something like this would be better:

/* This is used by vivin:gradle-semantic-build-versioning plugin to generate versioned jar files. */

PR #207, opened 2020-01-31
CommitLatencyMetrics and Corresponding Initial Unit Testing

andrewchoi5:
CommitLatencyMetrics and Corresponding Initial Unit Testing

Tested by locally deploying with zookeeper and kafka server instance. 
(Note, the implementation for latency metrics is in a separate PR)
image


smccauliff:
Have you run the code changes from the last PR within LinkedIn to see if it works when deployed?

andrewchoi5:
> Have you run the code changes from the last PR within LinkedIn to see if it works when deployed?

I will prioritize doing that, sure.

src/main/java/com/linkedin/kmf/services/CommitAvailabilityMetrics.java
```diff
@@ -47,8 +47,10 @@
 
     metrics.addMetric(new MetricName("offsets-committed-avg", METRIC_GROUP_NAME, "The average offset commits availability.", tags),
       (MetricConfig config, long now) -> {
-        double offsetsCommittedCount = metrics.metrics().get(metrics.metricName("offsets-committed-total", METRIC_GROUP_NAME, tags)).value();
-        double offsetsCommittedErrorCount = metrics.metrics().get(metrics.metricName("failed-commit-offsets-total", METRIC_GROUP_NAME, tags)).value();
+        double offsetsCommittedCount =
+            (double) metrics.metrics().get(metrics.metricName("offsets-committed-total", METRIC_GROUP_NAME, tags)).metricValue();
+        double offsetsCommittedErrorCount =
+            (double) metrics.metrics().get(metrics.metricName("failed-commit-offsets-total", METRIC_GROUP_NAME, tags)).metricValue();
```
smccauliff: Is this cast to double needed?

src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java
```diff
@@ -0,0 +1,56 @@
...
+
+
+/**
+ * The CommitLatencyMetrics class contains methods that measures and
+ * determines the latency of Kafka consumer offset commit().
+ */
+public class CommitLatencyMetrics {
+  private static final String METRIC_GROUP_NAME = "commit-latency-service";
+  private static final Logger LOG = LoggerFactory.getLogger(CommitLatencyMetrics.class);
+  final Sensor _commitOffsetLatency;
```
smccauliff: Why is this package private?

src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * Metrics for Calculating the offset commit latency of a consumer.
+   * @param metrics the commit offset metrics
+   * @param tags the tags associated, i.e) kmf.services:name=single-cluster-monitor
+   */
+  CommitLatencyMetrics(final Metrics metrics, final Map<String, String> tags,
+      final int latencyPercentileMaxMs,
+      final int latencyPercentileGranularityMs) {
```
smccauliff: All this excessive final just makes the code less readable and does not really help correctness.

src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java
```diff
@@ -0,0 +1,56 @@
...
+    LOG.info("{} was called.", this.getClass().getSimpleName());
```
smccauliff: Maybe put this at the end of the constructor with "{} was constructed successfully"

src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java
```diff
@@ -0,0 +1,56 @@
...
+
+    _commitOffsetLatency = metrics.sensor("commit-offset-latency");
+    _commitOffsetLatency.add(new MetricName("commit-offset-latency-ms-avg", METRIC_GROUP_NAME, "The average latency in ms of committing offset", tags), new Avg());
+    _commitOffsetLatency.add(new MetricName("commit-offset-latency-ms-max", METRIC_GROUP_NAME, "The maximum latency in ms of committing offset", tags), new Max());
+
+    int bucketNum = latencyPercentileMaxMs / latencyPercentileGranularityMs + 2;
+    int sizeInBytes = bucketNum * 4;
+    _commitOffsetLatency.add(new Percentiles(sizeInBytes, latencyPercentileMaxMs, Percentiles.BucketSizing.CONSTANT,
```
smccauliff: Is a way to compute these that does not rely on fixed sized histograms?

src/test/java/com/linkedin/kmf/services/ConsumeServiceTest.java
```diff
@@ -106,6 +106,37 @@ public void commitAvailabilityTest() throws Exception {
...
+
+    /* Thread.sleep safe to do here instead of ScheduledExecutorService
+     *  We want to sleep current thread so that consumeService can start running for enough seconds. */
+    Thread.sleep(threadStartDelay);
+
+    consumeService.stop();
+    consumeService.stop();
+
+    consumeService.awaitShutdown();
+  }
```
smccauliff: Doesn't all this consume service start and stop code already exist somewhere?  Perhaps refactor that code do it can be reused in other tests.

src/main/java/com/linkedin/kmf/services/CommitAvailabilityMetrics.java
andrewchoi5: Yes as the metricValue method returns an object.
```    public Object metricValue() {
        long now = time.milliseconds();
        synchronized (this.lock) {
            if (this.metricValueProvider instanceof Measurable)
                return ((Measurable) metricValueProvider).measure(config, now);
            else if (this.metricValueProvider instanceof Gauge)
                return ((Gauge<?>) metricValueProvider).value(config, now);
            else
                throw new IllegalStateException("Not a valid metric: " + this.metricValueProvider.getClass());
        }
    }```

File: src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java (hunk 2 above)
Comment (andrewchoi5) [reply src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java]: privatized.

File: src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java (hunk 3 above)
Comment (andrewchoi5) [reply src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java]: Sure. Amended.

File: src/test/java/com/linkedin/kmf/services/ConsumeServiceTest.java (hunk 6 above)
Comment (andrewchoi5) [reply src/test/java/com/linkedin/kmf/services/ConsumeServiceTest.java]: Sure. Will revise

File: src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java (hunk 4 above)
Comment (andrewchoi5) [reply src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java]: Sure.

File: src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java (hunk 5 above)
Comment (andrewchoi5) [reply src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java]: Unless we should deviate from how this repository initializes the Sensor objects for metrics, I believe it's a feasible option. 

For example, `ProduceService.java` class and `ConsumeMetrics` for `ConsumeService` class both have their `Sensor` objects which utilize the ../kafka/common/metrics/Percentiles class from kafka-clients package. 

The Percentiles constructor takes in ```int sizeInBytes, double max, BucketSizing bucketing, Percentile... percentiles)```.  I'm happy to accept suggestions.

File: src/main/java/com/linkedin/kmf/services/CommitAvailabilityMetrics.java (hunk 1 above)
Comment (smccauliff) [reply src/main/java/com/linkedin/kmf/services/CommitAvailabilityMetrics.java]: Can it return null?  If so then casting the return value to a primitive type will cause NPE.

File: src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java (hunk 5 above)
Comment (smccauliff) [reply src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java]: Never mind.  It's a battle for another day.

File: src/test/java/com/linkedin/kmf/services/ConsumeServiceTest.java [hunk 7]
```diff
@@ -184,4 +198,36 @@ public void run() {
...
+  /**
+   * return consume service metrics.
+   * set up the tags for the metrics
+   * @param consumeService ConsumeService object
+   * @return consume service metrics
+   */
+  private Metrics consumeServiceMetrics(ConsumeService consumeService) {
+    Metrics metrics = consumeService.metrics();
+    tags = new HashMap<>();
+    tags.put(TAGS_NAME, TAG_NAME_VALUE);
```
smccauliff: This could be put in the constructor of the test or by having a setup method like this:
```java
@BeforeMethod
public void setup() {
  tags = new HashMap<>();
  tags.put(TAGS_NAME, TAG_NAME_VALUE);
}
```

andrewchoi5: Amended.

PR #213, opened 2020-02-20
(work in progress, do not merge, do not review) Consumer Commit Offset Latency Implementation

andrewchoi5:
(work in progress, do not merge) Consumer Commit Offset Latency Implementation

### DIFF
--- a/src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java
+++ b/src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java
@@ -1,0 +1,56 @@
+/**
+ * Copyright 2020 LinkedIn Corp. Licensed under the Apache License, Version 2.0 (the "License"); you may not use this
+ * file except in compliance with the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package com.linkedin.kmf.services;
+
+import java.util.Map;
+import org.apache.kafka.common.MetricName;
+import org.apache.kafka.common.metrics.Metrics;
+import org.apache.kafka.common.metrics.Sensor;
+import org.apache.kafka.common.metrics.stats.Avg;
+import org.apache.kafka.common.metrics.stats.Max;
+import org.apache.kafka.common.metrics.stats.Percentile;
+import org.apache.kafka.common.metrics.stats.Percentiles;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * The CommitLatencyMetrics class contains methods that measures and
+ * determines the latency of Kafka consumer offset commit().
+ */
+public class CommitLatencyMetrics {
+  private static final String METRIC_GROUP_NAME = "commit-latency-service";
+  private static final Logger LOG = LoggerFactory.getLogger(CommitLatencyMetrics.class);
+  final Sensor _commitOffsetLatency;
+
+  /**
+   * Metrics for Calculating the offset commit latency of a consumer.
+   * @param metrics the commit offset metrics
+   * @param tags the tags associated, i.e) kmf.services:name=single-cluster-monitor
+   */
+  CommitLatencyMetrics(final Metrics metrics, final Map<String, String> tags,
+      final int latencyPercentileMaxMs,
+      final int latencyPercentileGranularityMs) {
+    LOG.info("{} was called.", this.getClass().getSimpleName());
+
+    _commitOffsetLatency = metrics.sensor("commit-offset-latency");
+    _commitOffsetLatency.add(new MetricName("commit-offset-latency-ms-avg", METRIC_GROUP_NAME, "The average latency in ms of committing offset", tags), new Avg());
+    _commitOffsetLatency.add(new MetricName("commit-offset-latency-ms-max", METRIC_GROUP_NAME, "The maximum latency in ms of committing offset", tags), new Max());
+
+    int bucketNum = latencyPercentileMaxMs / latencyPercentileGranularityMs + 2;
+    int sizeInBytes = bucketNum * 4;
+    _commitOffsetLatency.add(new Percentiles(sizeInBytes, latencyPercentileMaxMs, Percentiles.BucketSizing.CONSTANT,
+        new Percentile(new MetricName("commit-offset-latency-ms-99th", METRIC_GROUP_NAME, "The 99th percentile latency of committing offset", tags), 99.0),
+        new Percentile(new MetricName("commit-offset-latency-ms-999th", METRIC_GROUP_NAME, "The 99.9th percentile latency of committing offset", tags), 99.9),
+        new Percentile(new MetricName("commit-offset-latency-ms-9999th", METRIC_GROUP_NAME, "The 99.99th percentile latency of committing offset", tags), 99.99)));
+
+  }
+}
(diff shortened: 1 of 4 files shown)


src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java
```diff
@@ -0,0 +1,63 @@
...
+    int bucketNum = latencyPercentileMaxMs / latencyPercentileGranularityMs + 2;
+    int sizeInBytes = bucketNum * 4;
+    _commitOffsetLatency.add(new Percentiles(sizeInBytes, latencyPercentileMaxMs, Percentiles.BucketSizing.CONSTANT,
+        new Percentile(new MetricName("commit-offset-latency-ms-99th", METRIC_GROUP_NAME, "The 99th percentile latency of committing offset", tags), 99.0),
+        new Percentile(new MetricName("commit-offset-latency-ms-999th", METRIC_GROUP_NAME, "The 99.9th percentile latency of committing offset", tags), 99.9),
+        new Percentile(new MetricName("commit-offset-latency-ms-9999th", METRIC_GROUP_NAME, "The 99.99th percentile latency of committing offset", tags), 99.99)));
+
+  }
+
+  public void setCommittedMs(long time) {
```
Lincong: Since there is already a setter for `_committedMs`, maybe we can consider adding a getter for this field and make `_committedMs` private. This suggestion is more on the code style side

src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * The CommitLatencyMetrics class contains methods that measures and
+ * determines the latency of Kafka consumer offset commit().
+ */
+public class CommitLatencyMetrics {
+  private static final String METRIC_GROUP_NAME = "commit-latency-service";
+  private static final Logger LOG = LoggerFactory.getLogger(CommitLatencyMetrics.class);
+  final Sensor _commitOffsetLatency;
+  long _committedMs;
```
Lincong: Would `_committedStartTimeMs` be a better name than `_committedMs`?

src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java
```diff
@@ -0,0 +1,63 @@
...
+import org.apache.kafka.common.metrics.stats.Percentiles;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * The CommitLatencyMetrics class contains methods that measures and
+ * determines the latency of Kafka consumer offset commit().
+ */
+public class CommitLatencyMetrics {
```
Lincong: Some general suggestion/recommendation for APIs on this class. I would consider to have APIs such as:

```java
class CommitLatencyMetrics {
    public void recordCommitStart();
    public void recordCommitComplete();
}
```

There are two advantages If APIs are designed this way

1. We don't have to expose the actually numerical start timestamp values to users and we don't have to make users to generate the complete time stamp and do the calculation to figure out how many milliseconds have passed since the commit start and etc. All users need to do is to record when commit starts and record again when commit finishes.

2. It would be easier to detect abnormal/incorrect situations since the `recordCommitStart()` and `recordCommitComplete()` are always expected to called alternatively with `recordCommitStart()` being expected to be called first.

src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java
```diff
@@ -0,0 +1,63 @@
...
+  private static final String METRIC_GROUP_NAME = "commit-latency-service";
+  private static final Logger LOG = LoggerFactory.getLogger(CommitLatencyMetrics.class);
+  final Sensor _commitOffsetLatency;
```
Lincong: Can you make this field private and provide a getter method for it? I think in Java, it is always a good practice to make fields private and provides getters and/or setters for them if necessary. The reasons are:

1. Make the class more testable since the class would be easier for mocking
2. We can control whether we return a reference to the original field (the real guts) or make a copy of the object and return a reference to the copied object.
3. It opens up opportunity for us to instrument the getter or setter code path. For example, how many times some getter of some fields has been called and etc.

src/test/java/com/linkedin/kmf/services/ConsumeServiceTest.java
```diff
@@ -106,6 +106,37 @@ public void commitAvailabilityTest() throws Exception {
...
+
+    Assert.assertNull(metrics.metrics().get(metrics.metricName("commit-offset-latency-ms-avg", METRIC_GROUP_NAME, tags)));
+    Assert.assertNull(metrics.metrics().get(metrics.metricName("commit-offset-latency-ms-max", METRIC_GROUP_NAME, tags)));
+
+    /* Should start */
+    consumeService.start();
+    Assert.assertTrue(consumeService.isRunning());
+
+    /* in milliseconds */
+    long threadStartDelay = 1000 * THREAD_START_DELAY;
```
Lincong: 1. Can you rename the variable from `THREAD_START_DELAY ` to `THREAD_START_DELAY_SECONDS`?

2. Can you rename the variable from `threadStartDelay` to `threadStartDelayMs`? I think it is in general a good idea to let the variable name reflect the time unit.

3. Can you use `TimeUnit.SECONDS.toMillis(THREAD_START_DELAY_SECONDS);` to do the conversion?

src/test/java/com/linkedin/kmf/services/ConsumeServiceTest.java
```diff
@@ -106,6 +106,37 @@ public void commitAvailabilityTest() throws Exception {
...
+
+    /* Thread.sleep safe to do here instead of ScheduledExecutorService
+     *  We want to sleep current thread so that consumeService can start running for enough seconds. */
+    Thread.sleep(threadStartDelay);
+
+    consumeService.stop();
+    consumeService.stop();
```
Lincong: Do we need to stop the consumer service twice?

src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java
```diff
@@ -0,0 +1,63 @@
...
+        new Percentile(new MetricName("commit-offset-latency-ms-99th", METRIC_GROUP_NAME, "The 99th percentile latency of committing offset", tags), 99.0),
+        new Percentile(new MetricName("commit-offset-latency-ms-999th", METRIC_GROUP_NAME, "The 99.9th percentile latency of committing offset", tags), 99.9),
+        new Percentile(new MetricName("commit-offset-latency-ms-9999th", METRIC_GROUP_NAME, "The 99.99th percentile latency of committing offset", tags), 99.99)));
+
+  }
+
+  public void setCommittedMs(long time) {
+    _committedMs = time;
+  }
+
```
Lincong: Extra white space

src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java
```diff
@@ -0,0 +1,63 @@
...
+                       final Map<String, String> tags,
+                       final int latencyPercentileMaxMs,
+                       final int latencyPercentileGranularityMs) {
+    LOG.info("{} was called.", this.getClass().getSimpleName());
+
+    _commitOffsetLatency = metrics.sensor("commit-offset-latency");
+    _commitOffsetLatency.add(new MetricName("commit-offset-latency-ms-avg", METRIC_GROUP_NAME, "The average latency in ms of committing offset", tags), new Avg());
+    _commitOffsetLatency.add(new MetricName("commit-offset-latency-ms-max", METRIC_GROUP_NAME, "The maximum latency in ms of committing offset", tags), new Max());
+
+    int bucketNum = latencyPercentileMaxMs / latencyPercentileGranularityMs + 2;
```
Lincong: Can you check if `latencyPercentileGranularityMs` is 0 before the division? I understand that it may never be the case. However, I personally do favor a slightly more defensive style.

andrewchoi5:
@Lincong Hi Lincong, thank you for the review. However, as stated in the Title and Description of this PR, this pull request is "`work in progress`, `do ready to merge`, `not ready for review`".

andrewchoi5:
Thus, I will close this PR to avoid any confusion.

PR #214, opened 2020-02-21
Simplifications of CompletableFuture + Ensuring MultiClusterTopicManagementService performs the topic partitions operations first.

andrewchoi5:
This removes the cases where the Xinfra Monitor fails to locate the topic partitions for a topic. 
1 - Removal of Race Condition between `MultiClusterTopicManagementService` and `ProduceService/ConsumeService`
2 - Thread.sleep to ensure the `MultiClusterTopicManagementService` executes the operation of preparing topics (and creation of it if necessary), followed by the topic partitions.
3 - move the `topic-partitions-count` to `start()` ConsumeService such that the topic and partitions preparation is fully executed first by the `MultiClusterTopicManagementService` and `TopicManagementService` -> see (*1)
4 - Simplify the usage of  `CompletableFuture<Void>` by using only one of  `CompletableFuture<Void>` instead of two  `CompletableFuture<Void>`

(*1)
``` Sensor topicPartitionCount = metrics.sensor("topic-partitions");
      DescribeTopicsResult describeTopicsResult = _adminClient.describeTopics(Collections.singleton(_topic));
      Map<String, KafkaFuture<TopicDescription>> topicResultValues = describeTopicsResult.values();
      KafkaFuture<TopicDescription> topicDescriptionKafkaFuture = topicResultValues.get(_topic);
      TopicDescription topicDescription = null;
      try {
        topicDescription = topicDescriptionKafkaFuture.get();
      } catch (InterruptedException | ExecutionException e) {
        LOG.error("Exception occurred while getting the topicDescriptionKafkaFuture", e);
      }
      int partitionCount = topicDescription.partitions().size();
      topicPartitionCount.add(
          new MetricName("topic-partitions-count", METRIC_GROUP_NAME, "The total number of partitions for the topic.", tags),
          new CumulativeSum(partitionCount));```

Sample log output:

```2020-02-20 20:17:15 INFO  MultiClusterTopicManagementService:126 - single-cluster-monitor/MultiClusterTopicManagementService started.
2020-02-20 20:17:15 INFO  MultiClusterTopicManagementService:274 - CreateTopicsResult: {kafka-monitor-topic-z49=KafkaFuture{value=null,exception=null,done=false}}.
2020-02-20 20:17:15 INFO  Log4jControllerRegistration$:31 - Registered kafka:type=kafka.Log4jController MBean
2020-02-20 20:17:15 INFO  Metadata:261 - [Producer clientId=kmf-client-id] Cluster ID: 4mUaHr6GRk2B9NGkEwrxzg
```

```2020-02-20 20:17:20 INFO  DefaultMetricsReporterService:86 - ==============================================================
kmf:type=kafka-monitor:offline-runnable-count=0.0
kmf.services:name=single-cluster-monitor,type=produce-service:produce-availability-avg=1.0
kmf.services:name=single-cluster-monitor,type=consume-service:consume-availability-avg=1.0
kmf.services:name=single-cluster-monitor,type=commit-availability-service:offsets-committed-avg=1.0
kmf.services:name=single-cluster-monitor,type=commit-availability-service:failed-commit-offsets-avg=0.0
kmf.services:name=single-cluster-monitor,type=produce-service:records-produced-total=56.0
kmf.services:name=single-cluster-monitor,type=consume-service:records-consumed-total=38.0
kmf.services:name=single-cluster-monitor,type=consume-service:records-lost-total=0.0
kmf.services:name=single-cluster-monitor,type=consume-service:records-lost-rate=0.0
kmf.services:name=single-cluster-monitor,type=consume-service:records-duplicated-total=0.0
kmf.services:name=single-cluster-monitor,type=consume-service:records-delay-ms-avg=4.157894736842105
kmf.services:name=single-cluster-monitor,type=produce-service:records-produced-rate=0.9349070936075727
kmf.services:name=single-cluster-monitor,type=produce-service:produce-error-rate=0.0
kmf.services:name=single-cluster-monitor,type=consume-service:consume-error-rate=0.0
kmf.services:name=single-cluster-monitor,type=commit-availability-service:offsets-committed-total=1.0
kmf.services:name=single-cluster-monitor,type=commit-availability-service:failed-commit-offsets-total=0.0```

### DIFF
--- a/src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
+++ b/src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
@@ -31,6 +31,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
 import net.sourceforge.argparse4j.ArgumentParsers;
 import net.sourceforge.argparse4j.inf.ArgumentParser;
 import net.sourceforge.argparse4j.inf.Namespace;
@@ -71,16 +72,16 @@ public class SingleClusterMonitor implements App {
   }
 
   @Override
-  public void start() {
+  public void start() throws Exception {
     _topicManagementService.start();
     CompletableFuture<Void> topicPartitionResult = _topicManagementService.topicPartitionResult();
     try {
       /* Delay 2 second to reduce the chance that produce and consumer thread has race condition
       with TopicManagementService and MultiClusterTopicManagementService */
-      long threadSleepMs = 1000 * 2;
+      long threadSleepMs = TimeUnit.SECONDS.toMillis(2);
       Thread.sleep(threadSleepMs);
     } catch (InterruptedException e) {
-      LOG.error("InterruptedException while thread sleeping.", e);
+      throw new Exception("Interrupted while sleeping the thread", e);
     }
     CompletableFuture<Void> topicPartitionFuture = topicPartitionResult.thenRun(() -> {
       _produceService.start();
@@ -90,7 +91,7 @@ public class SingleClusterMonitor implements App {
     try {
       topicPartitionFuture.get();
     } catch (InterruptedException | ExecutionException e) {
-      LOG.error("Exception occurred while getting the TopicPartitionFuture", e);
+      throw new Exception("Exception occurred while getting the TopicPartitionFuture", e);
     }
 
     LOG.info(_name + "/SingleClusterMonitor started.");
--- a/src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
+++ b/src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
@@ -11,6 +11,7 @@
 package com.linkedin.kmf.apps;
 
 import com.linkedin.kmf.services.ConsumeService;
+import com.linkedin.kmf.services.ConsumerFactory;
 import com.linkedin.kmf.services.ConsumerFactoryImpl;
 import com.linkedin.kmf.services.DefaultMetricsReporterService;
 import com.linkedin.kmf.services.JettyService;
@@ -29,6 +30,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import net.sourceforge.argparse4j.ArgumentParsers;
 import net.sourceforge.argparse4j.inf.ArgumentParser;
 import net.sourceforge.argparse4j.inf.Namespace;
@@ -48,6 +50,7 @@ import org.slf4j.LoggerFactory;
 public class SingleClusterMonitor implements App {
   private static final Logger LOG = LoggerFactory.getLogger(SingleClusterMonitor.class);
 
+  private static final int SERVICES_INITIAL_CAPACITY = 4;
   private final TopicManagementService _topicManagementService;
   private final ProduceService _produceService;
   private final ConsumeService _consumeService;
@@ -55,14 +58,13 @@ public class SingleClusterMonitor implements App {
   private final List<Service> _allServices;
 
   public SingleClusterMonitor(Map<String, Object> props, String name) throws Exception {
+    ConsumerFactory consumerFactory = new ConsumerFactoryImpl(props);
     _name = name;
     _topicManagementService = new TopicManagementService(props, name);
-    CompletableFuture<Void> topicPartitionReady = _topicManagementService.topicPartitionResult();
+    CompletableFuture<Void> topicPartitionResult = _topicManagementService.topicPartitionResult();
     _produceService = new ProduceService(props, name);
-    ConsumerFactoryImpl consumerFactory = new ConsumerFactoryImpl(props);
-    _consumeService = new ConsumeService(name, topicPartitionReady, consumerFactory);
-    int servicesInitialCapacity = 4;
-    _allServices = new ArrayList<>(servicesInitialCapacity);
+    _consumeService = new ConsumeService(name, topicPartitionResult, consumerFactory);
+    _allServices = new ArrayList<>(SERVICES_INITIAL_CAPACITY);
     _allServices.add(_topicManagementService);
     _allServices.add(_produceService);
     _allServices.add(_consumeService);
@@ -71,11 +73,26 @@ public class SingleClusterMonitor implements App {
   @Override
   public void start() {
     _topicManagementService.start();
-    CompletableFuture<Void> completableFuture = _topicManagementService.topicManagementResult();
-    completableFuture.thenRun(() -> {
+    CompletableFuture<Void> topicPartitionResult = _topicManagementService.topicPartitionResult();
+    try {
+      /* Delay 2 second to reduce the chance that produce and consumer thread has race condition
+      with TopicManagementService and MultiClusterTopicManagementService */
+      long threadSleepMs = 1000 * 2;
+      Thread.sleep(threadSleepMs);
+    } catch (InterruptedException e) {
+      LOG.error("InterruptedException while thread sleeping.", e);
+    }
+    CompletableFuture<Void> topicPartitionFuture = topicPartitionResult.thenRun(() -> {
       _produceService.start();
       _consumeService.start();
     });
+
+    try {
+      topicPartitionFuture.get();
+    } catch (InterruptedException | ExecutionException e) {
+      LOG.error("Exception occurred while getting the TopicPartitionFuture", e);
+    }
+
     LOG.info(_name + "/SingleClusterMonitor started.");
   }
 
--- a/src/main/java/com/linkedin/kmf/services/ConsumerFactoryImpl.java
+++ b/src/main/java/com/linkedin/kmf/services/ConsumerFactoryImpl.java
@@ -18,10 +18,13 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Random;
+import java.util.concurrent.ExecutionException;
 import org.apache.kafka.clients.admin.AdminClient;
 import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.common.config.ConfigException;
 import org.apache.kafka.common.serialization.StringDeserializer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 
 public class ConsumerFactoryImpl implements ConsumerFactory {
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

File: src/main/java/com/linkedin/kmf/services/ConsumerFactoryImpl.java [hunk 1]
```diff
@@ -34,9 +37,16 @@
       new String[] {ConsumeServiceConfig.BOOTSTRAP_SERVERS_CONFIG, ConsumeServiceConfig.ZOOKEEPER_CONNECT_CONFIG};
   private int _latencySlaMs;
   private static AdminClient adminClient;
+  private static final Logger LOGGER = LoggerFactory.getLogger(ConsumerFactoryImpl.class);
```
Lincong: I think the Logger variable name is `LOG` instead of `LOGGER` through out this codebase. In most other codebases, the logger variable name is `LOG` instead of `LOGGER` as far as I have seen

src/main/java/com/linkedin/kmf/services/ConsumerFactoryImpl.java
```diff
@@ -34,9 +37,16 @@
...
 
   public ConsumerFactoryImpl(Map<String, Object> props) throws Exception {
+    LOGGER.info("Creating AdminClient.");
     adminClient = AdminClient.create(props);
+    try {
```
Lincong: I think it's in general a good idea to test adminClient in the constructor of this class. However, in the catch statement, if the `ExecutionException` exception was thrown, that indicates some issue with the admin client or the Kafka cluster, right? For example,

1. The admin client is not configured properly
2. The Kafka cluster has some ongoing issues so that metadata can not be returned
3. The network connect between admin client and the Kafka cluster has problems

So, is it a good idea to just log the exception and move on? Because we can potentially return a broken admin client to users of this factory class. Can we consider throw the exception in the constructor?

src/main/java/com/linkedin/kmf/services/ConsumerFactoryImpl.java
```diff
@@ -34,9 +37,16 @@
       new String[] {ConsumeServiceConfig.BOOTSTRAP_SERVERS_CONFIG, ConsumeServiceConfig.ZOOKEEPER_CONNECT_CONFIG};
   private int _latencySlaMs;
   private static AdminClient adminClient;
```
Lincong: 1. Can we name this field `_ adminClient` instead of `adminClient` so that naming conventions are consistent across all fields?
2. Can we add `final` to this field unless there is some reason not to do so?

src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
```diff
@@ -71,11 +73,26 @@ public SingleClusterMonitor(Map<String, Object> props, String name) throws Excep
   @Override
   public void start() {
     _topicManagementService.start();
-    CompletableFuture<Void> completableFuture = _topicManagementService.topicManagementResult();
-    completableFuture.thenRun(() -> {
+    CompletableFuture<Void> topicPartitionResult = _topicManagementService.topicPartitionResult();
```
Lincong: I found the naming of the 2 CompletableFuture variables used here to be a little bit confusing. If I understand what the code is supposed to be doing correctly, I would rename `topicPartitionResult` to `topicPartitionResultReady` and `topicPartitionFuture` to `producerConsumerReady`. The `ready` part is just my personal style. But making two variable names have the prefix `topicPartition` is just a little unintuitive to me

src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
```diff
@@ -71,11 +73,26 @@ public SingleClusterMonitor(Map<String, Object> props, String name) throws Excep
...
+    }
+    CompletableFuture<Void> topicPartitionFuture = topicPartitionResult.thenRun(() -> {
       _produceService.start();
       _consumeService.start();
     });
+
+    try {
+      topicPartitionFuture.get();
+    } catch (InterruptedException | ExecutionException e) {
+      LOG.error("Exception occurred while getting the TopicPartitionFuture", e);
```
Lincong: It's similar to my another comment on the `ConsumerFactoryImpl` constructor. If there is exception being thrown while waiting on the the producerService and consumer service to start, we should probably report/throw the exception to the caller to indicate that there is some problem occurred instead of just log the exception and then log `/SingleClusterMonitor started.` since it might not have started at all.

src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
```diff
@@ -71,11 +73,26 @@ public SingleClusterMonitor(Map<String, Object> props, String name) throws Excep
   @Override
   public void start() {
     _topicManagementService.start();
-    CompletableFuture<Void> completableFuture = _topicManagementService.topicManagementResult();
-    completableFuture.thenRun(() -> {
+    CompletableFuture<Void> topicPartitionResult = _topicManagementService.topicPartitionResult();
+    try {
+      /* Delay 2 second to reduce the chance that produce and consumer thread has race condition
+      with TopicManagementService and MultiClusterTopicManagementService */
+      long threadSleepMs = 1000 * 2;
```
Lincong: Can you use `TimeUnit.SECONDS.toMillis(2);`

src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
```diff
@@ -71,11 +73,26 @@ public SingleClusterMonitor(Map<String, Object> props, String name) throws Excep
   @Override
   public void start() {
     _topicManagementService.start();
-    CompletableFuture<Void> completableFuture = _topicManagementService.topicManagementResult();
-    completableFuture.thenRun(() -> {
+    CompletableFuture<Void> topicPartitionResult = _topicManagementService.topicPartitionResult();
+    try {
+      /* Delay 2 second to reduce the chance that produce and consumer thread has race condition
```
Lincong: I think this approach of using a hardcoded delay to prevent potential race condition is kind of uncommon or not the best practice in my opinion. In general, concurrency problems should NOT just be "reduce the chance" instead it should be solved/prevented completely by using concurrency control primitives (such as CompletableFuture in this case). So I need some clarification to see the purpose of this change.

Isn't the purpose of using the CompletableFuture to establish orders of service starting orders in order to prevent race conditions? I was reading through the code and I don't see why using CompletableFuture is not enough in this case or even if there is some race condition problem, why can't it solved theoretically by using the CompletableFuture.

src/main/java/com/linkedin/kmf/services/ConsumerFactoryImpl.java
andrewchoi5: Done

src/main/java/com/linkedin/kmf/services/ConsumerFactoryImpl.java
andrewchoi5: Sure.

src/main/java/com/linkedin/kmf/services/ConsumerFactoryImpl.java
andrewchoi5: 1. We should not do that. class-level static, non-final variables follow a naming convention where the underscore doesn't prefix the variable name. checkstyle is that is should match pattern '^[a-z][a-zA-Z0-9]*$'. [StaticVariableName] 
2. We also should not do that. That would force KMF to initialize the AdminClient in the class level declaration and not in the constructor. Let's create AdminClient object in the constructor.

src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
andrewchoi5: Done.

src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
andrewchoi5: We can throw Exception and add Exception signature to the App-level start() method.

src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
andrewchoi5: Please see the PR description bullet number (4): `4 - Simplify the usage of CompletableFuture<Void> by using only one of CompletableFuture<Void> instead of two CompletableFuture<Void>` We are getting rid of one of the Futures to simplify the understanding and make operationally more efficient. In the final product, we are only using Future for TopicPartition being ready in time.

src/main/java/com/linkedin/kmf/apps/SingleClusterMonitor.java
andrewchoi5: I've discussed this with Sean in the past in this regards. This a similar practice followed by Dong Lin when the ConsumeService was being implemented, in the `private void consume()` method. The TopicManagementService instantiates a new `MultiClusterTopicManagementService` in the constructor, which spins off a new executor thread (this is a separate thread). While the topicPartitionResult waits for the `MultiClusterTopicManagementService` to finish, the new scheduled thread is separate. It's been tested that giving a two second window for the the SingleClusterMonitor to wait is enough time frame to avoid race condition, as is originally done in the `consume()` of `ConsumeService`

```_executor = Executors.newSingleThreadScheduledExecutor(
      r -> new Thread(r, _serviceName + "-multi-cluster-topic-management-service"));```

File: src/main/java/com/linkedin/kmf/services/ConsumerFactoryImpl.java (hunk 3 above)
Comment (andrewchoi5) [reply src/main/java/com/linkedin/kmf/services/ConsumerFactoryImpl.java]: Please refer to https://iwww.corp.linkedin.com/wiki/cf/display/ENGS/Java+Coding+Conventions

Comment (Lincong) [reply src/main/java/com/linkedin/kmf/services/ConsumerFactoryImpl.java]: I see. I did not notice that the  AdminClient is a static field

Comment (andrewchoi5) [reply src/main/java/com/linkedin/kmf/services/ConsumerFactoryImpl.java]: No problem

**Comment by smccauliff** [comment]
I didn't do much work on the multi-cluster part of the monitor which is used to monitor mirroring pipelines.  Is there a case where different components of the app are deployed in different networks?  Are the producer and the consumer supposed to be running in the same process?  Can you check how we have this deployed at LinkedIn?

**Comment by andrewchoi5** [comment]
> I didn't do much work on the multi-cluster part of the monitor which is used to monitor mirroring pipelines. Is there a case where different components of the app are deployed in different networks? Are the producer and the consumer supposed to be running in the same process? Can you check how we have this deployed at LinkedIn?

``` Runnable tmRunnable = new TopicManagementRunnable();
      _executor.scheduleWithFixedDelay(tmRunnable, 0, _scheduleIntervalMs, TimeUnit.MILLISECONDS);
      Runnable pleRunnable = new PreferredLeaderElectionRunnable();
      _executor.scheduleWithFixedDelay(pleRunnable, _preferredLeaderElectionIntervalMs, _preferredLeaderElectionIntervalMs,
          TimeUnit.MILLISECONDS);
```
We use `Executors` with fixed scheduled with Runnables for the TopicManagement and for the PreferredLeaderElection (PLERunnable)

The producer instance and consumer instance are run in separate, independent threads for the `MultiClusterMonitor`.

andrewchoi5:
For our `public class PipelineDestinationClusterMonitor implements App` at Linkedin, we actually use  SingleClusterMonitor to monitor the  availability of a pipeline. They should be configured to use the same topic name and partition number. `PipelineDestinationClusterMonitor` beings a `ConsumeService` and `TopicManagementService` and expose metrics such as consume-availability-avg and records-delay-ms-avg etc.....

```@Override
  public void start() {
    _topicManagementService.start();
    _consumeService.start();
    LOG.info(_name + "/PipelineDestinationClusterMonitor started");
  }```

PR #219, opened 2020-02-28
Wake up consumers

andrewchoi5:
Wake up consumers gracefully


src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -194,6 +195,7 @@ public void onComplete(Map<TopicPartition, OffsetAndMetadata> topicPartitionOffs
       }
     }
     /* end of consume() while loop */
+    _baseConsumer.close();
```
andrewchoi5: graceful shutdown by closing when the boolean flag is set.

src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -61,6 +61,7 @@
   private String _name;
   private static final String METRIC_GROUP_NAME = "consume-service";
   private static Map<String, String> tags;
+  private final AtomicBoolean _closed = new AtomicBoolean(false);
```
Lincong: Why not just use the `_running` AtomicBoolean variable? When `awaitShutdown` or `stop` is called, set the `_running` value to be false, then the while loop in the `consume` method will exit.

In general, I prefer not to use multiple (atomic) Boolean variables together to represent states. One way that I aways use to represent a state machine (states + transitions among states) is to define an Enum and use AtomicReference<EnumDefined>

andrewchoi5: `_running` utilized.

src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -235,7 +237,7 @@ public synchronized void start() {
   public synchronized void stop() {
     if (_running.compareAndSet(true, false)) {
       try {
-        _baseConsumer.close();
+        _running.set(false);
```
smccauliff: This won't do anything since _running.compareAndSet would have already set this to false.

src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -245,6 +247,8 @@ public synchronized void stop() {
 
   @Override
   public void awaitShutdown() {
+    _running.set(false);
+    _baseConsumer.wakeup();
```
smccauliff: From the Service interface documentation:

```java
  /**
   * Implementations of this method must be thread safe and must be blocking.
   */
  void awaitShutdown();
```

PR #220, opened 2020-03-02
Remove unneeded dependencies espicaly on Jetty.

smccauliff:
Remove some unneeded dependencies espicaly on Jetty.  Jetty is only needed for
    the JolokiaService, but it will start it's on embedded webserver should someone
    want to run that service.
    Narrow the scope of the testng dependency.


smccauliff:
I  tested that KM still works, locally but we don't have any testing for any of the metrics systems that people have contributed.  Next time someone wants to modify that code they will need to provide some kind of test we can run to ensure we have not introduced regressions.

andrewchoi5:
LGTM.

sunnythepatel:
Hi Team,
  If jetty service has been removed how and what port I can check the UI on?

THank you

PR #225, opened 2020-03-05
Measurement of Commit Latency Service Metrics 

andrewchoi5:
Measurement of `Commit Latency Service` Metrics 
1 - Adds `setCommitStartTimeMs()` functionality
2 - Adds `public void recordCommitComplete()` for finishing the recording of consumer offset commit.
3 - Implements `  public void setCommitStartTimeMs(long time)` for setting in milliseconds the start time of consumer offset commit.
4 - Implements  `  public long commitStartTimeMs() ` for retrieving the start time of consumer offset commit.


src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java
```diff
@@ -41,14 +42,47 @@
     _commitOffsetLatency.add(new MetricName("commit-offset-latency-ms-avg", METRIC_GROUP_NAME, "The average latency in ms of committing offset", tags), new Avg());
     _commitOffsetLatency.add(new MetricName("commit-offset-latency-ms-max", METRIC_GROUP_NAME, "The maximum latency in ms of committing offset", tags), new Max());
 
+    if (latencyPercentileGranularityMs == 0) {
+      throw new IllegalArgumentException();
```
Lincong: Can you pass a String message describing the IllegalArgumentException to the exception constructor

src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java
```diff
@@ -41,14 +42,47 @@
...
+    }
     int bucketNum = latencyPercentileMaxMs / latencyPercentileGranularityMs + 2;
```
Lincong: Can you add a comment here explaining why `+ 2` ?

src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java
```diff
@@ -41,14 +42,47 @@
...
         new Percentile(new MetricName("commit-offset-latency-ms-999th", METRIC_GROUP_NAME, "The 99.9th percentile latency of committing offset", tags), 99.9),
         new Percentile(new MetricName("commit-offset-latency-ms-9999th", METRIC_GROUP_NAME, "The 99.99th percentile latency of committing offset", tags), 99.99)));
-
     LOG.info("{} was constructed successfully.", this.getClass().getSimpleName());
+  }
+
+  /**
+   * start the recording of consumer offset commit
+   */
+  public void recordCommitStart() {
```
Lincong: I would suggest you consider adding some defensive mechanisms to make sure the `recordCommitStart ()` is always called before `recordCommitComplete()` is called. I prefer this kind of defensive programming mindset especially when methods/APIs are expected to be called/invoked in a particular order. 

This defensive mindset assumes that users of the class (or the caller) might not behave as expected. In this case, that means not calling `recordCommitStart ()`, and `recordCommitComplete()` in an interleaving way. In general, it helps with debugging if something goes wrong.

src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java
andrewchoi5: sgtm

src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java
andrewchoi5: Done

src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java
andrewchoi5: Let's implement a defensive mechanisms ensuring `recordCommitStart()` is called before `recordCommitComplete()` is trigggered.

src/main/java/com/linkedin/kmf/services/CommitLatencyMetrics.java
```diff
@@ -30,25 +30,76 @@
   private static final String METRIC_GROUP_NAME = "commit-latency-service";
   private static final Logger LOG = LoggerFactory.getLogger(CommitLatencyMetrics.class);
   private final Sensor _commitOffsetLatency;
+  private long _commitStartTimeMs;
+  private static boolean inProgressCommit;
```
Lincong: 3 things.

1. Why does the variable name not start with `_` (`_ inProgressCommit` instead of `inProgressCommit`)

2. Why is this variable defined as `static`?

3. I think this variable might need to be defined as a volatile boolean variable. Since `recordCommitStart` and `recordCommitComplete` are called in different threads (IIUC). `recordCommitStart` is called in the `consume()` method where as `recordCommitComplete` is called in a callback.

Maybe you already know this. But in general, if multiple threads could change and read the state of some shared variable, that shared variable needs to be defined as volatile.
See https://www.logicbig.com/tutorials/core-java-tutorial/java-multi-threading/visibility-volatile.html

andrewchoi5: 1 - Because `inProgressCommit` is a private static boolean variable and not a private non-static boolean variable.
2 - Let's use non-static since we don't want to provide class level access to a this variable.
3 - sgtm

PR #231, opened 2020-04-02
Performance Improvements with Collection.singletonList

andrewchoi5:
Performance Improvements with Collection.singletonList

1 - use `Collections.singletonList("kmf.services:*:*")`list for performance improvements
2 - Redundant assignments removal.
3 - Use MetricValue casted to double. 

### DIFF
--- a/src/main/java/com/linkedin/kmf/services/ConsumeMetrics.java
+++ b/src/main/java/com/linkedin/kmf/services/ConsumeMetrics.java
@@ -77,7 +77,7 @@ public class ConsumeMetrics {
 
     metrics.addMetric(new MetricName("consume-availability-avg", METRIC_GROUP_NAME, "The average consume availability", tags),
       (config, now) -> {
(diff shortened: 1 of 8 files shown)


andrewchoi5:
> How does this improve performance?

@smccauliff 
1 - calls to Arrays.asList() with only one argument can be replaced with a call to Collections.singletonList() which will save memory overhead. Also the list returned by singletonList() is immutable whereas the list returned asList() allows calling the set() method. 

2 - putAll() calls after instantiation of a collection using a constructor call without arguments. This can be replaced with a single call to a parametrized constructor which is faster and performant

smccauliff:
The number of bytes saved is probably on the order of 12 bytes per config instance.  We might have 10-20 instances in production, so 240 bytes.  In production we have heap sizes with multiple Gbytes so .000004% savings in memory.

My point in not approving was that there are no substantive changes here.  But when we need to debug things in production someone will need to look at these changes to see if they broke something wasting everyone's time.

smccauliff:
I don't understand the comment about putAll() since it's not in this diff anywhere.

PR #232, opened 2020-04-02
MetricValue casted to double to avoid deprecated APIs.

andrewchoi5:
MetricValue casted to double to avoid deprecated APIs.

```        double recordsDelayedRate = (double) metrics.metrics().get(metrics.metricName("records-delayed-rate", METRIC_GROUP_NAME, tags)).metricValue();```


andrewchoi5:
> I have no specific concerns with this PR, other than not knowing what `.value()` returned, and what `.metricValue()` returns.

@mgrubent 
so `.value()` simply returns double. Currently, it's deprecated and is suggesting users to use metricValue. `metricValue` is a general type of Object. so we cast it do `double` as it's the type we used before.

andrewchoi5:
> Is there a plan to make a new release to support this?

@smccauliff 
Yes there will be new release to include this change in 2.1.5

PR #245, opened 2020-05-23
Xinfra Monitor Consumer Group Coordinator Assignment

andrewchoi5:
Xinfra Monitor Consumer Group Coordinator Assignment.
```
1. Implement Consumer Group Coordinator Assignment.
2. Bump up dependencies on Kafka and Kafka Client
```

### DIFF
--- a/src/main/java/com/linkedin/kmf/consumer/NewConsumer.java
+++ b/src/main/java/com/linkedin/kmf/consumer/NewConsumer.java
@@ -15,33 +15,86 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Properties;
+import java.util.concurrent.ExecutionException;
+import org.apache.kafka.clients.admin.AdminClient;
+import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.kafka.clients.consumer.OffsetAndMetadata;
 import org.apache.kafka.clients.consumer.OffsetCommitCallback;
 import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.common.internals.Topic;
+import org.apache.kafka.common.utils.Utils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-/*
- * Wrap around the new consumer from Apache Kafka and implement the #KMBaseConsumer interface
+
+/**
+ * Wraps around the new consumer from Apache Kafka and implement the #KMBaseConsumer interface
  */
 public class NewConsumer implements KMBaseConsumer {
 
   private final KafkaConsumer<String, String> _consumer;
   private Iterator<ConsumerRecord<String, String>> _recordIter;
-  private static final Logger LOG = LoggerFactory.getLogger(NewConsumer.class);
+  private static final Logger LOGGER = LoggerFactory.getLogger(NewConsumer.class);
   private static long lastCommitted;
 
-  public NewConsumer(String topic, Properties consumerProperties) {
+  public NewConsumer(String topic, Properties consumerProperties, AdminClient adminClient)
+      throws ExecutionException, InterruptedException {
+    LOGGER.info("{} is being instantiated in the constructor..", this.getClass());
+    this.configureGroupId(consumerProperties, adminClient);
     _consumer = new KafkaConsumer<>(consumerProperties);
     _consumer.subscribe(Collections.singletonList(topic));
   }
 
+  /**
+   * https://docs.confluent.io/current/clients/producer.html
+   * https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/producer/internals/DefaultPartitioner.java#L69
+   * The partitioner will hash the key with murmur2 algorithm
+   * and divide it by the number of partitions.
+   * The result is that the same key is always assigned to the same partition.
+   * @param groupId kafka consumer group ID
+   * @param consumerOffsetsTopicPartitions number of partitions in the __consumer_offsets topic.
+   * @return hashed integer
+   */
+  protected int groupIdHelper(String groupId, int consumerOffsetsTopicPartitions) {
+    // TODO: use long hash = UUID.nameUUIDFromBytes(s.getBytes()).getMostSignificantBits() ?
+    // TODO: MessageDigest digest = MessageDigest.getInstance("SHA-256");
+    // TODO: byte[] hash = digest.digest(text.getBytes(StandardCharsets.UTF_8));
+    return Utils.murmur2(groupId.getBytes()) % (consumerOffsetsTopicPartitions);
+  }
+
+  /**
+   * hash(group.id) % (number of __consumer_offsets topic partitions).
+   * The partition's leader is the group coordinator
+   * Choose B s.t hash(A) % (number of __consumer_offsets topic partitions) == hash(B) % (number of __consumer_offsets topic partitions)
+   * @param consumerProperties persistent set of kafka consumer properties
+   */
+  protected void configureGroupId(Properties consumerProperties, AdminClient adminClient)
+      throws ExecutionException, InterruptedException {
+    String consumerOffsetsTopic = Topic.GROUP_METADATA_TOPIC_NAME;
+    int numConsumerOffsetsTopicPartitions = adminClient.describeTopics(Collections.singleton(consumerOffsetsTopic))
+        .values()
+        .get(consumerOffsetsTopic)
+        .get()
+        .partitions()
+        .size();
+
+    String consumerGroupPrefix = "__shadow_group";
+    int counter = 0;
+    String groupId = "";
+    while (groupIdHelper(consumerGroupPrefix + counter, numConsumerOffsetsTopicPartitions) != groupIdHelper(groupId,
+        numConsumerOffsetsTopicPartitions)) {
+      counter++;
+    }
+    consumerProperties.put(ConsumerConfig.GROUP_ID_CONFIG, groupId);
+  }
+
   @Override
   public BaseConsumerRecord receive() {
-    if (_recordIter == null || !_recordIter.hasNext())
+    if (_recordIter == null || !_recordIter.hasNext()) {
       _recordIter = _consumer.poll(Duration.ofMillis(Long.MAX_VALUE)).iterator();
+    }
 
     ConsumerRecord<String, String> record = _recordIter.next();
     return new BaseConsumerRecord(record.topic(), record.partition(), record.offset(), record.key(), record.value());
@@ -81,5 +134,4 @@ public class NewConsumer implements KMBaseConsumer {
   public void updateLastCommit() {
     lastCommitted = System.currentTimeMillis();
   }
-
 }
--- a/src/main/java/com/linkedin/kmf/consumer/NewConsumer.java
+++ b/src/main/java/com/linkedin/kmf/consumer/NewConsumer.java
@@ -57,6 +57,8 @@ public class NewConsumer implements KMBaseConsumer {
   }
 
   /**
+   * https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/coordinator/group/GroupMetadataManager.scala#L189
+   * The consumer group string hash is used for this modulo operation.
    * https://docs.confluent.io/current/clients/producer.html
    * The partitioner will hash the key with murmur2 algorithm
    * and divide it by the number of partitions.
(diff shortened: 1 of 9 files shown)


src/main/java/com/linkedin/kmf/consumer/NewConsumer.java
```diff
@@ -15,33 +15,99 @@
...
+   * https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/producer/internals/DefaultPartitioner.java#L69
+   * @param groupId kafka consumer group ID
+   * @param consumerOffsetsTopicPartitions number of partitions in the __consumer_offsets topic.
+   * @return hashed integer which represents a number, the Math.abs value of which is the broker
+   * ID of the group coordinator, or the leader of the offsets topic partition.
+   */
+  protected int consumerGroupCoordinatorHasher(String groupId, int consumerOffsetsTopicPartitions) {
+    // TODO: use long hash = UUID.nameUUIDFromBytes(s.getBytes()).getMostSignificantBits() ?
+    // TODO: MessageDigest digest = MessageDigest.getInstance("SHA-256");
+    // TODO: byte[] hash = digest.digest(text.getBytes(StandardCharsets.UTF_8));
```
smccauliff, line 74: What is this?

src/main/java/com/linkedin/kmf/consumer/NewConsumer.java
```diff
@@ -15,33 +15,99 @@
...
+
+    LOGGER.debug("Hashed and modulo output: {}", groupId.hashCode());
+    return Math.abs(groupId.hashCode()) % consumerOffsetsTopicPartitions;
```
smccauliff, line 77: The Kafka Utils.abs is slightly different:
```java
 /**
     * Get the absolute value of the given number. If the number is Int.MinValue return 0. This is different from
     * java.lang.Math.abs or scala.math.abs in that they return Int.MinValue (!).
     */
    public static int abs(int n) {
        return (n == Integer.MIN_VALUE) ? 0 : Math.abs(n);
    }
```

src/main/java/com/linkedin/kmf/consumer/NewConsumer.java
```diff
@@ -15,33 +15,99 @@
...
+   * The partitioner will hash the key with murmur2 algorithm
+   * and divide it by the number of partitions.
+   * The result is that the same key is always assigned to the same partition.
+   * https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/producer/internals/DefaultPartitioner.java#L69
+   * @param groupId kafka consumer group ID
+   * @param consumerOffsetsTopicPartitions number of partitions in the __consumer_offsets topic.
+   * @return hashed integer which represents a number, the Math.abs value of which is the broker
+   * ID of the group coordinator, or the leader of the offsets topic partition.
+   */
+  protected int consumerGroupCoordinatorHasher(String groupId, int consumerOffsetsTopicPartitions) {
```
smccauliff, line 71: Why does this have a different name than the original function in the Kafka code base?

src/main/java/com/linkedin/kmf/consumer/NewConsumer.java
```diff
@@ -15,33 +15,99 @@
...
+   * Choose B s.t hash(A) % (number of __consumer_offsets topic partitions) == hash(B) % (number of __consumer_offsets topic partitions)
+   * @param consumerProperties persistent set of kafka consumer properties
+   */
+  protected void configureGroupId(Properties consumerProperties, AdminClient adminClient)
+      throws ExecutionException, InterruptedException {
+    if (_targetConsumerGroupId.equals("")) {
+      throw new IllegalArgumentException("The target consumer group identifier cannot be empty: " + _targetConsumerGroupId);
+    }
+
+    int numOffsetsTopicPartitions = adminClient.describeTopics(Collections.singleton(Topic.GROUP_METADATA_TOPIC_NAME))
```
smccauliff, line 92: Is AdminClient needed since Consumer.partitionsFor() should also return the list of partitions for a topic?

src/main/java/com/linkedin/kmf/consumer/NewConsumer.java
```diff
@@ -15,33 +15,99 @@
...
 public class NewConsumer implements KMBaseConsumer {
 
   private final KafkaConsumer<String, String> _consumer;
   private Iterator<ConsumerRecord<String, String>> _recordIter;
-  private static final Logger LOG = LoggerFactory.getLogger(NewConsumer.class);
+  private static final Logger LOGGER = LoggerFactory.getLogger(NewConsumer.class);
   private static long lastCommitted;
+  protected String _targetConsumerGroupId;
+  protected String _consumerGroupPrefix = "__shadow_consumer_group-";
+  protected int _consumerGroupSuffixCandidate = 0;
```
smccauliff, line 42: Do these need to be data members a not local variables?   Why is_consumerGroupPrefix mutable?

src/main/java/com/linkedin/kmf/services/MultiClusterTopicManagementService.java
```diff
@@ -466,9 +467,10 @@ private static void reassignPartitions(KafkaZkClient zkClient, Collection<Node>
       }
 
       scala.collection.immutable.Set<String> topicList = new scala.collection.immutable.Set.Set1<>(topic);
-      scala.collection.Map<Object, scala.collection.Seq<Object>> currentAssignment = zkClient.getPartitionAssignmentForTopics(topicList).apply(topic);
-      String currentAssignmentJson = formatAsReassignmentJson(topic, currentAssignment);
-      String newAssignmentJson = formatAsReassignmentJson(topic, assignedReplicas);
+      scala.collection.Map<Object, ReplicaAssignment>
+          currentAssignment = zkClient.getPartitionAssignmentForTopics(topicList).apply(topic);
+      String currentAssignmentJson = formatAsOldAssignmentJson(topic, currentAssignment);
+      String newAssignmentJson = formatAsNewReassignmentJson(topic, assignedReplicas);
```
smccauliff, line 473: This is for the kafka version bump?

src/main/java/com/linkedin/kmf/consumer/NewConsumer.java
andrewchoi5: Either AdminClient or KafkaConsumer works.

src/main/java/com/linkedin/kmf/consumer/NewConsumer.java
```diff
@@ -15,33 +15,99 @@
...
+      this.configureGroupId(consumerProperties, adminClient);
+    }
     _consumer = new KafkaConsumer<>(consumerProperties);
     _consumer.subscribe(Collections.singletonList(topic));
   }
 
+  /**
+   * https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/coordinator/group/GroupMetadataManager.scala#L189
+   * The consumer group string hash is used for this modulo operation.
+   * https://docs.confluent.io/current/clients/producer.html
```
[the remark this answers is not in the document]
andrewchoi5: removed

src/main/java/com/linkedin/kmf/consumer/NewConsumer.java
andrewchoi5: removed

src/main/java/com/linkedin/kmf/consumer/NewConsumerConfig.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+public class NewConsumerConfig extends AbstractConfig {
+
+  private static final ConfigDef CONFIG_DEF;
+
+  public static final String TARGET_CONSUMER_GROUP_ID_CONFIG = "target.consumer.group.id";
+  public static final String TARGET_CONSUMER_GROUP_ID_CONFIG_DOC =
+      "The identifier (in String) of the consumer group in which the target consumer - whose group coordinator to which "
+          + "Xinfra Monitor consumer should commit its consumer offsets is responsible for receiving its committed "
+          + "offsets - resides.";
```
smccauliff, line 28: I find this hard to read.  Does this work for you?

"When defined a consumer group is chosen such that it maps to the same group coordinator as the specified group coordinator."

andrewchoi5: Works.

andrewchoi5: 👍

src/main/java/com/linkedin/kmf/services/ConsumeService.java
```diff
@@ -86,27 +86,35 @@ public ConsumeService(String name,
...
-        } catch (Exception e) {
-          LOG.error(name + "/ConsumeService failed", e);
-        }
-      }, name + " consume-service");
-      _consumeThread.setDaemon(true);
+    // Returns a new CompletionStage (topicPartitionFuture) which
+    // executes the given action - code inside run() - when this stage (topicPartitionResult) completes normally,.
+    CompletableFuture<Void> topicPartitionFuture = topicPartitionResult.thenRun(new Runnable() {
+      @Override
+      public void run() {
```
smccauliff, line 93: This is the old syntax why switch to this now?

src/main/java/com/linkedin/kmf/services/MultiClusterTopicManagementService.java
andrewchoi5: yes.

src/main/java/com/linkedin/kmf/consumer/NewConsumer.java
```diff
@@ -15,33 +15,94 @@
...
   private Iterator<ConsumerRecord<String, String>> _recordIter;
-  private static final Logger LOG = LoggerFactory.getLogger(NewConsumer.class);
+  private static final Logger LOGGER = LoggerFactory.getLogger(NewConsumer.class);
   private static long lastCommitted;
 
-  public NewConsumer(String topic, Properties consumerProperties) {
+  // package private permission for NewConsumerTest to access.
+  String _consumerGroupPrefix = "__shadow_consumer_group-";
+  int _consumerGroupSuffixCandidate = 0;
+  String _targetConsumerGroupId;
```
smccauliff, line 45: There's no reason for these to be instance variables.  These are never referred to after being used once.   PartitionsFor() doesn't need to be an instance member.   These choices make this functionality hard to test.

String _consumerGroupPrefix = "__shadow_consumer_group-";,  This never changes, it can be a constant. If it needs to be referred to for testing purposes then mark it as "static final".  If it's package private then tests can access it as well.

```java
class SomeUtilClass {
/** documentation */
static int partitionFor(String groupId, int consumerOffsetsTopicPartitions) {

    LOGGER.debug("Hashed and modulo output: {}", groupId.hashCode());
    return Utils.abs(groupId.hashCode()) % consumerOffsetsTopicPartitions;
  }

/** documentation */
  // Instead of making targetGroupId an instance variable and then assigning it some value which this then looks up
  // it can just be a parameter to a method
  static String findCollision(String targetGroupId, AdminClient adminClient)
      throws ExecutionException, InterruptedException {
    if (targetGroupId.equals("")) {
      throw new IllegalArgumentException("The target consumer group identifier cannot be empty: " + targetGroupId);
    }

 int numOffsetsTopicPartitions = adminClient.describeTopics(Collections.singleton(Topic.GROUP_METADATA_TOPIC_NAME))
	
        .values()
        .get(Topic.GROUP_METADATA_TOPIC_NAME)
        .get()
        .partitions()
        .size();

   // Extract invariant from loop
   int targetConsumerOffsetsPartition = partitionsFor(targetGroupId, numOffsetsTopicPartitions);

   //  This doesn't need to be an instance variable because we throw this out this value at the end of computation
    int groupSuffix = 0;

    // Extract return value so it's not computed twice, this reduces the possibility of bugs
    String newGroupName = null;

    // Use while(true) otherwise halting condition is hard to read.
    while (true) {
      //TODO:  could play fancy StringBuilder games here to make this generate less garbage
       newGroupName = GROUP_SUFFIX_CANDIDATE + Integer.toString(groupSuffix++);
       int newGroupNamePartition = partitionsFor(newGroupName, numOffsetsTopicPartitions);
        if (targetConsumerOffsetsPartition == newGroupNamePartition) {
           break;
        }
    }
   return newGroupName;
}

```

I ended up writing this code as a comment.  It probably doesn't compile.  It's here to convey ideas.
For testing I would just run partitionsFor() on the result to make sure they are the same and that they groupIds are not the same.

andrewchoi5: `       newGroupName = GROUP_SUFFIX_CANDIDATE + Integer.toString(groupSuffix++);` 
--> did you mean to put `GROUP_PREFIX_CANDIDATE`?

andrewchoi5: Other than that, I'm content with adopting the above class that's proposed.

PR #253, opened 2020-05-30
Fallback to regular constructor parameters if the constructor doesn't include AdminClient.

andrewchoi5:
Fallback to regular constructor parameters if the constructor doesn't include AdminClient.

### DIFF
--- a/src/main/java/com/linkedin/kmf/services/ConsumerFactoryImpl.java
+++ b/src/main/java/com/linkedin/kmf/services/ConsumerFactoryImpl.java
@@ -97,7 +97,7 @@ public class ConsumerFactoryImpl implements ConsumerFactory {
     try {
       return Class.forName(consumerClassName).getConstructor(String.class, Properties.class, AdminClient.class);
     } catch (java.lang.NoSuchMethodException noSuchMethodException) {
-      LOG.info("The method was not found. This is not a constructor with AdminClient parameter: ", noSuchMethodException);
+    LOG.info(consumerClassName + " does not provide a constructor with signature (Ljava/lang/String;Ljava/util/Properties;Lorg/apache/kafka/clients/admin/AdminClient;)V - falling back to (Ljava/util/Properties;)V");
       return null;
     } catch (ClassNotFoundException e) {
       throw new ClassNotFoundException("The class was not found: ", e);
--- a/src/main/java/com/linkedin/kmf/services/ConsumerFactoryImpl.java
+++ b/src/main/java/com/linkedin/kmf/services/ConsumerFactoryImpl.java
@@ -81,10 +81,27 @@ public class ConsumerFactoryImpl implements ConsumerFactory {
       props.forEach(consumerProps::putIfAbsent);
     }
 
-    _baseConsumer = (KMBaseConsumer) Class.forName(consumerClassName)
-        .getConstructor(String.class, Properties.class, AdminClient.class)
-        .newInstance(_topic, consumerProps, adminClient());
+    if (adminClientConstructorExists(consumerClassName)) {
+      _baseConsumer = (KMBaseConsumer) Class.forName(consumerClassName)
+          .getConstructor(String.class, Properties.class, AdminClient.class)
+          .newInstance(_topic, consumerProps, adminClient());
+    } else {
+      _baseConsumer = (KMBaseConsumer) Class.forName(consumerClassName)
+          .getConstructor(String.class, Properties.class)
+          .newInstance(_topic, consumerProps);
+    }
+  }
 
+  static boolean adminClientConstructorExists(String consumerClassName) throws ClassNotFoundException {
+    try {
+      Class.forName(consumerClassName).getConstructor(String.class, Properties.class, AdminClient.class);
+      return true;
+    } catch (java.lang.NoSuchMethodException noSuchMethodException) {
+      LOG.error("The method was not found. This is not a constructor with AdminClient parameter: ", noSuchMethodException);
+      return false;
+    } catch (ClassNotFoundException e) {
+      throw new ClassNotFoundException("The class was not found: ", e);
+    }
   }
 
   @Override
--- a/src/main/java/com/linkedin/kmf/services/ConsumerFactoryImpl.java
+++ b/src/main/java/com/linkedin/kmf/services/ConsumerFactoryImpl.java
@@ -81,9 +81,9 @@ public class ConsumerFactoryImpl implements ConsumerFactory {
       props.forEach(consumerProps::putIfAbsent);
     }
 
-    if (adminClientConstructorExists(consumerClassName)) {
-      _baseConsumer = (KMBaseConsumer) Class.forName(consumerClassName)
-          .getConstructor(String.class, Properties.class, AdminClient.class)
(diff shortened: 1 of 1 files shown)


src/main/java/com/linkedin/kmf/consumer/NewConsumerKMBaseConsumerFactoryImpl.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Random;
+import org.apache.kafka.clients.admin.AdminClient;
+import org.apache.kafka.clients.consumer.ConsumerConfig;
+import org.apache.kafka.common.config.ConfigException;
+import org.apache.kafka.common.serialization.StringDeserializer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class NewConsumerKMBaseConsumerFactoryImpl implements KMBaseConsumerFactory {
```
jjkoshy, line 27: Rather than use an altogether new factory impl, you could do this in the existing one using reflection - i.e., if `getConstructor(Map.class)` does not exist, then fall back to `getConstructor(String.class, Properties.class, AdminClient.class)`

That said, I don't really understand how this is intended to be used. i.e., what uses `ConsumerFactoryImpl` and what uses `NewConsumerKMBaseConsumerFactoryImpl` ?

Also, why does `ConsumerFactory` contain methods such as `latencySlaMs()`, etc.? Seems a bit weird to me.

I haven't been following all the KMF patches lately, so sorry about my ignorance on the current state.

andrewchoi5: Within ConsumeService, we reference these instance variables from `ConsumerFactory.`
```
_latencySlaMs = consumerFactory.latencySlaMs(); 
_adminClient = consumerFactory.adminClient();
 ....
consumerFactory.latencyPercentileMaxMs(),
consumerFactory.latencyPercentileGranularityMs());
_commitLatencyMetrics = new CommitLatencyMetrics(metrics, tags, consumerFactory.latencyPercentileMaxMs(),
```

src/main/java/com/linkedin/kmf/services/ConsumerFactoryImpl.java
```diff
@@ -81,10 +81,27 @@ public ConsumerFactoryImpl(Map<String, Object> props) throws Exception {
...
+          .newInstance(_topic, consumerProps);
+    }
+  }
 
+  static boolean adminClientConstructorExists(String consumerClassName) throws ClassNotFoundException {
+    try {
+      Class.forName(consumerClassName).getConstructor(String.class, Properties.class, AdminClient.class);
+      return true;
+    } catch (java.lang.NoSuchMethodException noSuchMethodException) {
+      LOG.error("The method was not found. This is not a constructor with AdminClient parameter: ", noSuchMethodException);
```
jjkoshy, line 100: I think this should be an info or debug log. i.e., it is not an error since you fall back to the properties-based constructor.

src/main/java/com/linkedin/kmf/services/ConsumerFactoryImpl.java
```diff
@@ -81,10 +81,27 @@ public ConsumerFactoryImpl(Map<String, Object> props) throws Exception {
...
+          .getConstructor(String.class, Properties.class, AdminClient.class)
+          .newInstance(_topic, consumerProps, adminClient());
+    } else {
+      _baseConsumer = (KMBaseConsumer) Class.forName(consumerClassName)
+          .getConstructor(String.class, Properties.class)
+          .newInstance(_topic, consumerProps);
+    }
+  }
 
+  static boolean adminClientConstructorExists(String consumerClassName) throws ClassNotFoundException {
```
jjkoshy, line 95: private

src/main/java/com/linkedin/kmf/services/ConsumerFactoryImpl.java
```diff
@@ -81,10 +81,27 @@ public ConsumerFactoryImpl(Map<String, Object> props) throws Exception {
       props.forEach(consumerProps::putIfAbsent);
     }
 
-    _baseConsumer = (KMBaseConsumer) Class.forName(consumerClassName)
-        .getConstructor(String.class, Properties.class, AdminClient.class)
-        .newInstance(_topic, consumerProps, adminClient());
+    if (adminClientConstructorExists(consumerClassName)) {
+      _baseConsumer = (KMBaseConsumer) Class.forName(consumerClassName)
+          .getConstructor(String.class, Properties.class, AdminClient.class)
```
jjkoshy, line 86: Rather than reflect twice (i.e., the new helper and here), maybe you can arrange to have the helper return the constructor if it exists, null otherwise?

src/main/java/com/linkedin/kmf/consumer/NewConsumerKMBaseConsumerFactoryImpl.java
jjkoshy: I see - it is slightly weird in that a `factory<t>` should ideally only deal with dispensing instances of `t`. Here it appears to be used to stash away some additional config information that is used elsewhere. It's NBD though, and orthogonal to what you are trying to do in this patch, so I'll just ignore that :)

src/main/java/com/linkedin/kmf/services/ConsumerFactoryImpl.java
```diff
@@ -81,10 +81,27 @@ public ConsumerFactoryImpl(Map<String, Object> props) throws Exception {
...
+          .newInstance(_topic, consumerProps);
+    }
+  }
 
+  private static java.lang.reflect.Constructor<?> adminClientConstructorIfExists(String consumerClassName)
+      throws ClassNotFoundException {
+    try {
+      return Class.forName(consumerClassName).getConstructor(String.class, Properties.class, AdminClient.class);
+    } catch (java.lang.NoSuchMethodException noSuchMethodException) {
+      LOG.info("The method was not found. This is not a constructor with AdminClient parameter: ", noSuchMethodException);
```
jjkoshy, line 100:
```suggestion
    LOG.info(consumerClassName + " does not provide a constructor with signature (Ljava/lang/String;Ljava/util/Properties;Lorg/apache/kafka/clients/admin/AdminClient;)V - falling back to (Ljava/util/Properties;)V");
```

andrewchoi5:
> (no need to show the exception in the message)
> I'm not sure what your plan is wrt deprecation - do you intend to always support these variants? And do you always intend to prefer the constructor variant with `AdminClient` over the one with `Properties` (since that is what the patch currently does)?

Good point. @jjkoshy 
Similar to how the Service Factory was implemented with the following PR, (https://github.com/linkedin/kafka-monitor/pull/257) we plan to have a consolidated consumer factories for KMF as well. 

For instance, we have a ConsumerFactoryImpl which gets referenced from `SingleClusterMonitor`, `MultClusterMonitor`, `BasicEndToEnd`, etc.

In such case, Xinfra Monitor would have NewConsumerFactoryImpl, (internal MP) RestConsumerFactoryImpl, KmLinkedinKafkaConsumerFactoryImpl.. etc b/c we already have a ConsumerFactory in open source KMF.

PR #258, opened 2020-06-01
Xinfra Monitor Rebranding Operation from Kafka Monitor

andrewchoi5:
`Xinfra Monitor` Rebranding Operation from Kafka Monitor

```Signed-off-by: Andrew Choi <li_andchoi@microsoft.com>```

### DIFF
--- a/bin/kafka-monitor-start.sh
+++ b/bin/kafka-monitor-start.sh
@@ -9,4 +9,4 @@
 
 base_dir=$(dirname $0)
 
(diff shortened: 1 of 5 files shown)


andrewchoi5:
> This seems to change some, but not all, of the Kafka Monitor language over to Xinfra Monitor.
> 
> The changes looks correct, but not complete.

That's correct. I plan to segment this rebranding operation into a few PRs.

PR #261, opened 2020-06-03
Remove deprecations in produce service, render fields local instead of instance variables

andrewchoi5:
Remove deprecations in produce service, render fields local instead of instance variables


andrewchoi5:
@mgrubent

PR #264, opened 2020-06-04
Conversion of Set to List for BrokerMetadata Collection

andrewchoi5:
Set is unordered and doesn't allow for duplicates.
List is ordered and allows for duplicates.

Converting Set -> List for O(1) runtime complexity in lieu of O(n) caused by the usage of List.

Set -> List should be safe in this case since the Set<brokerMetadata> contains non-duplicated values.

TopicFactory interface declares the following member functions. Changing this method to return List could break the backward compatibility
```Set<Integer> getBlackListedBrokers(String zkUrl);```


andrewchoi5:
@sutambe 
chose set -> list conversion since TopicFactory interface declares the following member functions. Changing this method to return List could break the backward compatibility..

`Set<Integer> getBlackListedBrokers(String zkUrl);`

PR #269, opened 2020-06-11
[WIP] Addition of JettyService enabling HTML webpage with real-time graphs

andrewchoi5:
1 - Addition of JettyService enabling HTML webpage with real-time graphs
2 - This PR is from a customer ask -- attached below.

```
..... watched a video series 'Apache Kafka Series - Kafka Monitoring
 and Operations' from 2018 that had a walk-through of how to set up Kafka Monitor.
 All he did was edit the single-cluster-monitor JSON in kafka-monitor.properties
 with the right ZK and bootstrap servers, 
then ran kafka-monitor-start.sh to start it (with the config), and 
metrics started flowing to the console. Like a miracle! Then he brought up the 
UI which is full of really useful information that's updated in real time.

I want to add it to our arsenal of observability tools ASAP but
 cannot get it working.

I tried the master branch first on our Kerberized Dev cluster
 (the config file and the command I used to start it are below) but it never brings up
 the app - I never see anything listening on port 
8000. It doesn't write any metrics to the console & 
doesn't bring up the webapp: it just seems to be waiting.

```


src/main/java/com/linkedin/kmf/services/JettyService.java
```diff
@@ -0,0 +1,74 @@
...
+import org.eclipse.jetty.server.Server;
+import org.eclipse.jetty.server.handler.ResourceHandler;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Jetty server which serves HTML files. (index.html)
+ */
+public class JettyService implements Service {
```
smccauliff, line 24: This is not the correct way to do this.  The correct way to do this is to create a war file and then deploy the war file into jetty.  No Java code is required, maybe some gradle code.

build.gradle
```diff
@@ -40,6 +40,7 @@ allprojects {
     compile 'com.signalfx.public:signalfx-codahale:0.0.47'
     compile group: 'org.apache.kafka', name: 'kafka_2.12', version: '2.3.1'
     compile group: 'org.apache.kafka', name: 'kafka-clients', version: '2.3.1'
+    compile 'org.eclipse.jetty:jetty-server:8.1.19.v20160209'
```
smccauliff, line 43: This version of Jetty has severe security vulnerabilities which is why it was removed in the first place.

config/xinfra-monitor.properties
```diff
@@ -69,6 +69,11 @@
     }
   },
 
+  "jetty-service": {
```
smccauliff, line 72: See comment below

smccauliff: Or above

smccauliff:
The problem with this kind of user request is that we don't use this feature within LinkedIn.   There's no domain expertise within LinkedIn on feature the user wants (in this case jolokia).  The feature can be hard to test or  doesn't have any tests.  Therefore it will not be well supported and my constitute tech debt from the first first commit.  There's a few options on how to deal with this:
1. Don't do it
2. Ask the user to create the PR and be the maintainer of this feature.
3. Ask the user to create the PR and create the tests

I'm open to other suggestions.

build.gradle
andrewchoi5: Upgraded to 
```
// https://mvnrepository.com/artifact/org.eclipse.jetty/jetty-server
compile group: 'org.eclipse.jetty', name: 'jetty-server', version: '9.4.29.v20200521'
```

src/main/java/com/linkedin/kmf/services/JettyService.java
andrewchoi5: I read your comment below -- as long as testing is in place, I think a web application + UI that displays real-time monitoring metrics is useful for open source users and more or less expected of Xinfra Monitor as shown in some of the series/walkthroughs such as the above. 

I will create an ask to the original requester to add `testing`.

A fourth option would be creating a separate branch that contains Jetty support for HTML and have the user commit to and maintain it.

config/xinfra-monitor.properties
andrewchoi5: commented on the first two reviews.

andrewchoi5:
I read your comment above -- I believe there's enough demand to support this in our repository. 
That being said, I agree that this should hold true so long as testing is in place; I think a web application + UI that displays real-time monitoring metrics is useful for open source users and more or less expected of Xinfra Monitor as shown in some of the series/walkthroughs such as the above.

I will create an ask to the original requester to maintain ownership on this part of the code or add some testing. (Ken and David from UltimateSoftware)

A fourth option would be creating a separate branch that contains Jetty support for HTML and have the user commit to and maintain it.

andrewchoi5:
Out of curiosity, is taking the servlet container approach first better? WAR or servlet containers are a pretty outdated approach to hosting HTTP servers. Embedded is also a simpler approach that's more commonly used.

Is the building a WAR file to deploy better than using an embed jetty server? 

Non-embedded way requires deploying jetty as a servlet container which means executing it as an independent application and then deploying the WAR file into it (by copying it into a certain directory). I guess that would mean it will be more orthogonal so they will be more decoupled.

andrewchoi5:
This PR is not from internal ask and from an external client's request. What would be the argument for keeping these two components decoupled? -- the application and this PR's new jetty server hosting `index.html` file. From my end, the only downside of using embedded jetty is that it may be more difficult to configure different jetty functionalities; however, because the `index.html` is actually straightforward and is not very complicated, there's not much configuration needed anyway.

PR #271, opened 2020-06-15
Monitoring Full Propagation of Topic in Cluster -- Cluster Topic Manipulation Service 

andrewchoi5:
1 - Monitoring Full Propagation of Topics within a Cluster -- Cluster Topic Manipulation Service. This service periodically creates and deletes topic and measures the duration from operation start to full propagation of the metadata to all the brokers. 

All Kafka brokers can answer a `topic metadata request` that describes the current state of the cluster, such as what topics there are, which partitions those topics have, which broker is the leader for those partitions, and the host and port information for these brokers.
The client finds a broker and that broker tells the client regarding all the other brokers which exist and what partitions they host. (This first broker may itself go down so the best practice for a client is to take ask two or three bootstrap brokers).
The topic creation could be issued through `AdminClient` and we could have a scenario where we send a request to broker A for metadata fetch. The broker could fetch the metadata and return a good signal and but in reality the broker B could be experiencing `extremely slow` speed/network and so it might not have even finished creating these topic partitions (`propagating metadata`).
At the point in time, checking the `log directories of broker B` will return negative (bad) since the topic partitions haven’t been created for that partition.

In other words, cluster topic metadata is eventually consistent. when metadata updates are propagated by the controller, the brokers can see them in any order and there can be delays. the client deals with this by getting `NO_TOPIC_OR_PARTITION` or other exception messages, which triggers it to do the lookup cycle over again, re-attempting until it gets a response from the broker that is supposed to be the leader.

If we send broker a fetch topic metadata request -> let’s say A says that the topic exists in the metadata.
it can be the case that it turns out broker B is extremely slow due to bad network or heavy queue and it doesn’t have the topic partitions populated in its log directory.

2 - TODO: Metrics/Sensors to record the time it takes to finish this metadata change (topic creation). -- separate PR
3 - TODO: Metrics/Sensors to record the time it takes to finish this metadata change (topic deletion). -- separate PR
4 - TODO in a sep. PR: make runnable interval configurable.
`Andrew Choi`


src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
```diff
@@ -0,0 +1,242 @@
...
+
+
+/**
+ * Service monitoring the creations and deletions of Kafka Cluster's Topic.
+ */
+public class ClusterTopicManipulationService implements Service {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(ClusterTopicManipulationService.class);
+  private final String _configDefinedServiceName;
+  private final int _reportIntervalSecond;
```
Lincong, line 41: Nit: I highly suggest using `java.time.Duration` whenever we need a variable to represent a time duration/interval

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
```diff
@@ -0,0 +1,242 @@
...
+    _executor = Executors.newSingleThreadScheduledExecutor();
+    _reportIntervalSecond = 1;
+    _configDefinedServiceName = name;
+    // TODO: instantiate a new instance of ClusterTopicManipulationMetrics(..) here.
+  }
+
+  /**
+   * The start logic must only execute once.  If an error occurs then the implementer of this class must assume that
+   * stop() will be called to clean up.  This method must be thread safe and must assume that stop() may be called
+   * concurrently. This can happen if the monitoring application's life cycle is being managed by a container.  Start
```
Lincong, line 65: It seems to me that the `start()` and `stop()` methods are not thread-safe and there is no guarantee that the `start logic must only execute once`. I would add two things.

1. Thread-safety to these 2 methods

2. Some states to keep track of how many times the `start()` method has been invoked and throw an IllegalStateException if any unexpected invocation is encountered

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
```diff
@@ -0,0 +1,242 @@
...
+        LOGGER.trace("{}-clusterTopicManipulationServiceRunnable successful!", this.getClass().getSimpleName());
+      }
+    } catch (InterruptedException | ExecutionException e) {
+      LOGGER.error("Exception occurred while creating cluster topic in {}: ", _configDefinedServiceName, e);
+    }
+  }
+
+  private boolean doesClusterContainTopic(String topic, Collection<Node> brokers, AdminClient adminClient)
+      throws ExecutionException, InterruptedException {
+    for (Node broker : brokers) {
```
Lincong, line 147: This for-loop is not necessary. The `LOGGER.trace()` can be merged to the below for-loop.
I guess that `adminClient.describeLogDirs` is expensive RPC. In this for-loop, the results from this RPC are ignored most of the time and the below for-loop does it again but actually uses the results. So, please merge these 2 for-loops and make sure `adminClient.describeLogDirs` is called once for every broker only.

src/main/java/com/linkedin/kmf/XinfraMonitorConstants.java
```diff
@@ -18,6 +18,13 @@
   public XinfraMonitorConstants() {
   }
 
+  public static final String TOPIC_MANIPULATION_SERVICE_TOPIC =
+      "xinfra-monitor-cluster-topic-manipulation-service-topic-";
+
+  public static final String KAFKA_LOG_DIRECTORY = "/tmp/kafka-logs";
+
+  public static final int TOPIC_MANIPULATION_TOPIC_NUM_PARTITIONS = 3;
```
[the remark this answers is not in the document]
andrewchoi5: Good point. This will be made configurable in future patch

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
andrewchoi5: 👍

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
andrewchoi5: 👍

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
```diff
@@ -0,0 +1,242 @@
...
+  public void start() {
+    LOGGER.info("ClusterTopicManipulationService started for {} - {}", _configDefinedServiceName,
+        this.getClass().getCanonicalName());
+    Runnable clusterTopicManipulationServiceRunnable = new ClusterTopicManipulationServiceRunnable();
+
+    _executor.scheduleAtFixedRate(clusterTopicManipulationServiceRunnable, _reportIntervalSecond, _reportIntervalSecond,
+        TimeUnit.SECONDS);
+  }
+
+  private class ClusterTopicManipulationServiceRunnable implements Runnable {
```
Lincong, line 84: I think you can just make `ClusterTopicManipulationService` a `Runnable` and in the start() method, submit `this` to the executor. No need to create another private inner class and the only purpose is to schedule a periodical invocation to the outer class's method

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
andrewchoi5: Updated the two methods to behave in a way that's similar to existing ProduceService and ConsumeService. This should suffice

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
```diff
@@ -0,0 +1,242 @@
...
+   */
+  @Override
+  public void awaitShutdown() {
+
+    try {
+      _executor.awaitTermination(3, TimeUnit.MINUTES);
+    } catch (InterruptedException e) {
+      LOGGER.info("Thread interrupted when waiting for {} to shutdown", _configDefinedServiceName);
+    }
+    LOGGER.info("{} shutdown completed", _configDefinedServiceName);
```
Lincong, line 233: I would place this line right after `_executor.awaitTermination(3, TimeUnit.MINUTES);`

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
```diff
@@ -0,0 +1,242 @@
...
+  public boolean isRunning() {
+
+    return !_executor.isShutdown();
+  }
+
+  /**
+   * Implementations of this method must be thread safe and must be blocking.
+   */
+  @Override
+  public void awaitShutdown() {
```
Lincong, line 226: I know this method is just implementing the interface method `public void awaitShutdown();` But I strongly encourage you to change the interface signature to let users pass in the timeout duration in the same way as the `awaitTermination(...)` on the executor interface.

Since `awaitShutdown()` without any timeout duration specified is just absurd to me and considering this repo is open-sourced, I really think we should make this change.

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
```diff
@@ -0,0 +1,242 @@
...
+          adminClient.describeLogDirs(Collections.singleton(broker.id())).all().get();
+
+      this.processBroker(logDirectoriesResponseMap, broker, topic);
+    }
+    return _totalPartitions.get() == 0;
+  }
+
+  void processBroker(Map<Integer, Map<String, DescribeLogDirsResponse.LogDirInfo>> logDirectoriesResponseMap,
+      Node broker, String topic) {
+    LOGGER.trace("logDirectoriesResponseMap: {}", logDirectoriesResponseMap);
```
Lincong, line 167: Can you confirm that in this case, the map content is going to be logged instead of just the map address?

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
Lincong: `adminClient.describeLogDirs(Collections.singleton(broker.id())).all().get()` is still invoked twice back to back right? I think even though `LOGGER.tracd(...)` usually does not log anything. But `adminClient.describeLogDirs(Collections.singleton(broker.id())).all().get()` is still going to be invoked there and it's just it's results are not logged.

You can call `adminClient.describeLogDirs(Collections.singleton(broker.id())).all().get()` only once and use its results for logging and later process

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
```diff
@@ -0,0 +1,248 @@
...
+          adminClient.describeLogDirs(Collections.singleton(broker.id())).all().get());
+      Map<Integer, Map<String, DescribeLogDirsResponse.LogDirInfo>> logDirectoriesResponseMap =
+          adminClient.describeLogDirs(Collections.singleton(broker.id())).all().get();
+
+      this.processBroker(logDirectoriesResponseMap, broker, topic);
+    }
+    return _totalPartitions.get() == 0;
+  }
+
+  void processBroker(Map<Integer, Map<String, DescribeLogDirsResponse.LogDirInfo>> logDirectoriesResponseMap,
```
Lincong, line 165: Two things.

1. Please add Java doc to describe the purpose of this method and how it achieves the purpose. It took me a while to understand that it's doing

2. Why is the `_totalPartitions ` needed? Can we just have `processBroker(...)` return the number of partitions it has discovered and sum them up. Then compare it with the expected total number of partitions?

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
```diff
@@ -0,0 +1,248 @@
...
+            _currentlyOngoingTopic);
+        _isOngoingTopicCreationDone = true;
+        LOGGER.trace("{}-clusterTopicManipulationServiceRunnable successful!", this.getClass().getSimpleName());
+      }
+    } catch (InterruptedException | ExecutionException e) {
+      LOGGER.error("Exception occurred while creating cluster topic in {}: ", _configDefinedServiceName, e);
+    }
+  }
+
+  private boolean doesClusterContainTopic(String topic, Collection<Node> brokers, AdminClient adminClient)
```
Lincong, line 151: Need Java doc to explain how it checks whether a cluster contains a topic or not. It is not very intuitive to understand.

Question: shouldn't we check whether a cluster contains a topic by fetching metadata instead of checking brokers' log directories? Because a topic does not exist until its metadata is in the results of the fetch metadata request. If KMF is measuring how long it takes to create and delete a topic from the user's PoV, it should look at the fetched metadata instead of the log dirs on Kafka brokers;

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
```diff
@@ -0,0 +1,248 @@
...
+      int random = ThreadLocalRandom.current().nextInt();
+      _currentlyOngoingTopic = XinfraMonitorConstants.TOPIC_MANIPULATION_SERVICE_TOPIC + Math.abs(random);
+
+      try {
+        int brokerCount = _adminClient.describeCluster().nodes().get().size();
+        _createTopicsResult = _adminClient.createTopics(Collections.singleton(
+            new NewTopic(_currentlyOngoingTopic, XinfraMonitorConstants.TOPIC_MANIPULATION_TOPIC_NUM_PARTITIONS,
+                (short) brokerCount)));
+
+        _totalPartitions.set(brokerCount * XinfraMonitorConstants.TOPIC_MANIPULATION_TOPIC_NUM_PARTITIONS);
```
Lincong, line 125: I actually don't understand this line. I thought `TOPIC_MANIPULATION_TOPIC_NUM_PARTITIONS ` is the number of partitions in the created topic. But `brokerCount * XinfraMonitorConstants.TOPIC_MANIPULATION_TOPIC_NUM_PARTITIONS` seems like we are expecting `TOPIC_MANIPULATION_TOPIC_NUM_PARTITIONS` partitions on every broker???

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
andrewchoi5: Yes.

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
andrewchoi5: let's leave it as a private class for clarity and also for leaving a room in case we will need to add additional functionalities to this Runnable in the future.

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
```diff
@@ -0,0 +1,248 @@
...
+      try {
+        ClusterTopicManipulationService.this.createDeleteClusterTopic();
+      } catch (Exception e) {
+        LOGGER.error("{} {} failed to run createDeleteClusterTopic()", _configDefinedServiceName,
+            ClusterTopicManipulationService.this.getClass().getSimpleName(), e);
+      }
+    }
+  }
+
+  private void createDeleteClusterTopic() {
```
Lincong, line 112: This method is probably the most important part of this PR (or this whole feature). Several things:
1. Please add Java doc to explain steps this method is taking. It took me a while to understand. In the Java doc, explicitly state whether this method is thread-safe or does it need to handle concurrent accesses.

2. IIUC, this method is thread-safe since you are using a `newSingleThreadScheduledExecutor` to schedule it. If that's true, what's the point of using a volatile boolean `_isOngoingTopicCreationDone` given that only one thread can access this method at a time?

3.  It seems that we are not waiting for topics to be created. Looking at `if (_createTopicsResult.all().isDone() && doesClusterContainTopic(_currentlyOngoingTopic, brokers,
          _adminClient))`, I think if a topic does not exist at this point, the `else` branch is executed. I thought we want to measure how long it takes to create a topic right? If there is no blocking and waiting for a topic to be created, how do we know how long it takes to create the topic?

Maybe my above questions are results of myself not understanding exactly what this method is doing. So, please clarify and add Java doc

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
andrewchoi5: It's true that metadata fetch is useful. Does AdminClient provide fetching metadata of brokers? 

If that was the case, rather than `instead` of checking brokers' log dirs, it should be fetching metadata `and` checking brokers' log directories. We should add metadata fetch if possible in the next PR.

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
andrewchoi5: `brokerCount * XinfraMonitorConstants.TOPIC_MANIPULATION_TOPIC_NUM_PARTITIONS)` is the total number of replicas (partitions) that exist for a topic. 
`TOPIC_MANIPULATION_TOPIC_NUM_PARTITIONS` is the number of partitions (not the RF)

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
andrewchoi5: JAVA doc was appended.

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
andrewchoi5: Added the JavaDoc.
3 - there is no else branch. If the topic partitions log directories aren't fully populated at that point, it will exit `createDeleteClusterTopic` method and in the next iteration of the `ClusterTopicManipulationServiceRunnable`, it will check again if the topic has been created across the brokers. 
I understand your point that we could just block. But the ClusterTopicManipulationServiceRunnable repeats and runs it self periodically at a non-trivial interval anyway to measure the latency at a granular level.

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
Lincong: If `TOPIC_MANIPULATION_TOPIC_NUM_PARTITIONS` is the number of partition for a topic, shouldn't the total number of replicas that exist for a topic be `TOPIC_MANIPULATION_TOPIC_NUM_PARTITIONS` * RF instead of `TOPIC_MANIPULATION_TOPIC_NUM_PARTITIONS` * brokerCount?

`TOPIC_MANIPULATION_TOPIC_NUM_PARTITIONS * brokerCount` seems that each broker has `TOPIC_MANIPULATION_TOPIC_NUM_PARTITIONS` partition.

I am confused about how is the total number of replicas associated with the broker count.

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
Lincong: I kind of understand your point. I think it's like `polling` vs `interrupt`. If we block and wait on the KafkaFuture, it's like `interrupt` whereas your approach is like `polling` (keep asking if something has happened).

I still think it's very awkward to use this `polling` method for 2 reasons.
1. We already have the opportunity to block and wait since we can wait on the KafkaFuture. In other words, Kafka topic creation/deletion APIs can let us block and wait. Why are we still going with the less accurate and less intuitive polling method? 

2. As you said `ClusterTopicManipulationServiceRunnable  repeats and runs itself periodically at a non-trivial interval`, that means the accuracy of the measure depends on the frequency at which the ClusterTopicManipulationServiceRunnable calls the method `createDeleteClusterTopic`. Whereas if we block and wait on the KafkaFuture, it is very accurate regarding how much time it took for the controller to create/delete a topic

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
Lincong: I quickly checked the `AdminClient` API. I don't think there is a method that lets us send a fetch metadata request to the Kafka server. Here are 2 things you can consider.

1. Use the `describeTopics` methods. The idea is to keep calling `describeTopics` to describe the topic you are trying to create until you see this method returns results for the created topic. Same thing for deleted topics. Keep describing deleted topics until you see it stops returning information about the deleted topics.

2. Construct a fetch metadata request yourself and send it to Kafka broker. I am not sure how hard it would be. But worth exploring.

Overall IMO, **checking whether brokers' log dirs indicate topic creation or topic deletion is almost meaningless (sorry for making it sound harsh) since that's NOT what users see.** We all know that there is some delay after log dirs created for a topic until users see the topic in metadata and then be able to use the topic. This delay might vary and definitely worth measuring. You can choose to implement fetching metadata in the next PR. But I do NOT agree that we should keep this approach and use it in production since it is meaningless.

What do you think @smccauliff ?

andrewchoi5:
Makes sense, I will leave it up to you to close. I resolved the comments after I have added clarifications/explanations and for the ones I deemed were finished.

For the implementations, let's chat offline on slack. It seems like we have concerns on polling/interrupt and TOPIC_MANIPULATION_TOPIC_NUM_PARTITIONS

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
andrewchoi5: The RF is set to the brokerCount currently to enable maximize spreading the partitions and replicas across all the brokers in the clusters as possible.
Feel free to disagree or suggest otherwise.

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
andrewchoi5: I have no disagreement nor a large affinity toward one solution.
If you think that the level of granularity needs to be more sensitive than it currently is, it's perfectly fine calling `.all().get()` on the KafkaFuture to block the call.
I will make that change.

Lincong: Sounds good. My whole point is that `calling .all().get() on the KafkaFuture to block the call` is easily available to us and will give us the most accurate measure. Plus it is how users of the AdminClient check whether their topics are created or deleted successfully. 

So There is no reason for not taking it.

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
Lincong: I think the below statement is totally valid and reasonable.
```
The RF is set to the brokerCount currently to enable maximize assigning the many partitions and replicas across all the brokers in the clusters as possible
```
Maybe we should document or annotate the code with this information? Maybe this information is documented somewhere in this codebase. But just looking at this part, I had a difficult time figuring out why `TOPIC_MANIPULATION_TOPIC_NUM_PARTITIONS * brokerCount`

Lincong:
> Makes sense, I will leave it up to you to close. I resolved the comments after I have added clarifications/explanations and for the ones I deemed were finished.
> 
> For the implementations, let's chat offline on slack. It seems like we have concerns on polling/interrupt and TOPIC_MANIPULATION_TOPIC_NUM_PARTITIONS

I think `brokerCount  * TOPIC_MANIPULATION_TOPIC_NUM_PARTITIONS` is reasonable and just need some documentation to make it clear that `brokerCount == RF`

Regarding polling/interrupt, we can chat offline. Please involve @smccauliff  in our discussion.

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
andrewchoi5: Added. documentation on it.

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
andrewchoi5: Okay to resolve this?

andrewchoi5:
S

> > Makes sense, I will leave it up to you to close. I resolved the comments after I have added clarifications/explanations and for the ones I deemed were finished.
> > For the implementations, let's chat offline on slack. It seems like we have concerns on polling/interrupt and TOPIC_MANIPULATION_TOPIC_NUM_PARTITIONS
> 
> I think `brokerCount * TOPIC_MANIPULATION_TOPIC_NUM_PARTITIONS` is reasonable and just need some documentation to make it clear that `brokerCount == RF`
> 
> Regarding polling/interrupt, we can chat offline. Please involve @smccauliff in our discussion.

Sure. 
Added documentations for the number of partitions and the replication factor for the topic.

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
andrewchoi5: Got back from lunch. I believe the remaining topic to be discussed is :

```I quickly checked the AdminClient API. I don't think there is a method that lets us send a fetch metadata request to the Kafka server. Here are 2 things you can consider.
Use the describeTopics methods. The idea is to keep calling describeTopics to describe the topic you are trying to create until you see this method returns results for the created topic. Same thing for deleted topics. Keep describing deleted topics until you see it stops returning information about the deleted topics.
Construct a fetch metadata request yourself and send it to Kafka broker. I am not sure how hard it would be. But worth exploring.```

As I commented right before your reply, `If that was the case, rather than instead of checking brokers' log dirs, it should be fetching metadata and checking brokers' log directories. We should add metadata fetch if possible in the next PR.` , could we include the the fetch metadata or the AdminClient describeTopics as part of the next PR?

I will not ELR this commit with the next release. I am happy to adopt one of the two approaches you suggested.

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
Lincong: OK

andrewchoi5:
As for feasibility of having both of these: 
1 - fetching the metadata or doing describeTopic and
2 - verifying that the log dirs have the topic partitions

All Kafka brokers can answer a topic metadata request that describes the current state of the cluster, such as what topics there are, which partitions those topics have, which broker is the leader for those partitions, and the host and port information for these brokers.
The client finds a broker and that broker tells the client regarding all the other brokers which exist and what partitions they host. (This first broker may itself go down so the best practice for a client is to take ask two or three bootstrap brokers).
The topic creation could be issued through AdminClient and we could have a scenario where we send a request to broker A for metadata fetch. The broker could fetch the metadata and return a good signal and but in reality the broker B could be experiencing extremely slow speed/network and so it might not have even finished creating these topic partitions (propagating metadata).
At the point in time, checking the log directories of broker B will return negative (bad) since the topic partitions haven’t been created for that partition.

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
andrewchoi5: As for feasibility of having both of these:
1 - fetching the metadata or doing describeTopic and
2 - verifying that the log dirs have the topic partitions

All Kafka brokers can answer a topic metadata request that describes the current state of the cluster, such as what topics there are, which partitions those topics have, which broker is the leader for those partitions, and the host and port information for these brokers.
The client finds a broker and that broker tells the client regarding all the other brokers which exist and what partitions they host. (This first broker may itself go down so the best practice for a client is to take ask two or three bootstrap brokers).
The topic creation could be issued through AdminClient and we could have a scenario where we send a request to broker A for metadata fetch. The broker could fetch the metadata and return a good signal and but in reality the broker B could be experiencing extremely slow speed/network and so it might not have even finished creating these topic partitions (propagating metadata).
At the point in time, checking the log directories of broker B will return negative (bad) since the topic partitions haven’t been created for that partition.

andrewchoi5:
In other words, cluster topic metadata is eventually consistent. when metadata updates are propagated by the controller, the brokers can see them in any order and there can be delays. the client deals with this by getting NO_TOPIC_OR_PARTITION or other exception messages, which triggers it to do the lookup cycle over again, re-attempting until it gets a response from the broker that is supposed to be the leader.

If we send broker a fetch topic metadata request -> let’s say A says that the topic exists in the metadata.
it can be the case that it turns out broker B is extremely slow due to bad network or heavy queue and it doesn’t have the topic partitions populated in its log directory.

PR #274, opened 2020-06-18
Implementation of whether the Topic Describe API is Successful or Not.

andrewchoi5:
Implementation of whether the Topic Describe API is Successful or Not.

### DIFF
--- a/src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
+++ b/src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
@@ -25,6 +25,8 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import org.apache.kafka.clients.admin.AdminClient;
 import org.apache.kafka.clients.admin.CreateTopicsResult;
 import org.apache.kafka.clients.admin.NewTopic;
+import org.apache.kafka.clients.admin.TopicDescription;
+import org.apache.kafka.common.KafkaFuture;
 import org.apache.kafka.common.Node;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.requests.DescribeLogDirsResponse;
@@ -157,8 +159,10 @@ public class ClusterTopicManipulationService implements Service {
    * @param brokers brokers to check log dirs from
    * @param adminClient Admin Client
    * @return true if the cluster contains the topic.
-   * @throws ExecutionException
-   * @throws InterruptedException
+   * @throws ExecutionException when attempting to retrieve the result of a task
+   * that aborted by throwing an exception.
+   * @throws InterruptedException when a thread is waiting, sleeping, or occupied,
+   * and the thread is interrupted, either before or during the activity.
    */
   private boolean doesClusterContainTopic(String topic, Collection<Node> brokers, AdminClient adminClient, int expected)
       throws ExecutionException, InterruptedException {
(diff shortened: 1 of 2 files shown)


src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
```diff
@@ -171,7 +175,28 @@ private boolean doesClusterContainTopic(String topic, Collection<Node> brokers,
 
       totalPartitionsInCluster += this.processBroker(logDirectoriesResponseMap, broker, topic);
     }
-    return totalPartitionsInCluster == expected;
+    return totalPartitionsInCluster == expected && ClusterTopicManipulationService.isTopicDescribeSuccessful(
```
Lincong, line 178: I should have pointed it out from the previous PR. Can you rename `expected` to `expectedTotalPartitionsInCluster` or something more descriptive?

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
```diff
@@ -171,7 +175,28 @@ private boolean doesClusterContainTopic(String topic, Collection<Node> brokers,
...
+  }
+
+  /**
+   * Waits if necessary for this future to complete, and then returns its result.
+   * returns true if the future succeeds, which occurs only if all the topic descriptions are successful.
+   * @param adminClient administrative client for Kafka, supporting managing and inspecting topics, brokers, configurations and ACLs.
+   * @param topicNames Collection of topic names
+   * @return boolean value if describe topic succeeds.
+   */
+  private static boolean isTopicDescribeSuccessful(AdminClient adminClient, Collection<String> topicNames) {
```
Lincong, line 189: This function does 2 things, waiting for the describe PRC to finish, check whether it failed and return a boolean. I would suggest the following:

```
private static Map<String, TopicDescription> describeTopics(AdminClient adminClient, Collection<String> topicNames) throws  InterruptedException, ExecutionException {
   KafkaFuture<Map<String, TopicDescription>> mapKafkaFuture = adminClient.describeTopics(topicNames).all();
   return mapKafkaFuture.get();
}
```

Then make the caller handle exception if there is any and decide whether topic description is successful or not. This way is cleaner since the function is doing only one thing instead of two things. You can also have the option to do something with the returned value if it's necessary in the future.

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
andrewchoi5: ack.

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
andrewchoi5: ack.

src/main/java/com/linkedin/kmf/services/ClusterTopicManipulationService.java
```diff
@@ -171,7 +175,37 @@ private boolean doesClusterContainTopic(String topic, Collection<Node> brokers,
 
       totalPartitionsInCluster += this.processBroker(logDirectoriesResponseMap, broker, topic);
     }
-    return totalPartitionsInCluster == expected;
+    return totalPartitionsInCluster == expectedTotalPartitionsInCluster
```
Lincong, line 178: Actually what I meant is:

```
if (totalPartitionInCluster != expected) {
    return false;
}

boolean describeSucceed = true;
try {
    Map<String, TopicDescription> topicDescriptions = ClusterTopicManipulationService.describeTopics(adminClient, topicNames);
    LOGGER.trace("topicDescriptionMap = {}", topicDescriptions);
} catch (InterruptedException | ExecutionException e) {
    describeSucceed = false;
    LOGGER.error("Exception occurred within describeTopicsFinished method for topics {}", topicNames, e);
}

return !describeSucceed;
```

In the above approach

1. We removed the `isTopicDescribeSuccessful` method. In your updated version, `isTopicDescribeSuccessful` is still doing two things, waiting and return boolean. You just factored out the waiting part to its own function. But that does not change the semantics to `isTopicDescribeSuccessful`

2. The code is very readable since readers would know that we wait and if results are returned successfully, we are good to go. Instead of looking at the `isTopicDescribeSuccessful` method, reading its comments and then realizing that this function blocks for results and log the exception if there is any, etc.

3. Reduce the number of static functions we have. There are some Java developers (I am not one of them) who are very against the idea of having static method in a class unless a class is used as a pure utility class. But I do agree that we should try to avoid unnecessary static functions.

andrewchoi5: thanks for your opinion

PR #280, opened 2020-06-18
Create maven.yml + This workflow will build a Java project with Maven

andrewchoi5:
This workflow will build a Java project with Maven. Gradle will be supported also.
https://github.com/linkedin/kafka-monitor/pull/278/files


smccauliff:
Choosing maven over gradle is also choosing to one of the only people at LinkedIn that use maven.

andrewchoi5:
Gradle will still be supported.

PR #292, opened 2020-07-02
Avoid NaN in Commit Offset Failure Rate OffsetCommitServiceMetrics by…

andrewchoi5:
… initializing with a zero only when instantiating.

**Avoid having XM reporting NaN for its failure rate by initializing the value of the failure rate sensor with a zero value.**

```
[2020-07-02 15:06:54,609] INFO ==============================================================
kmf:type=kafka-monitor:offline-runnable-count=0.0
kmf.services:name=offset-commit-service,type=offset-commit-service:offset-commit-availability-avg=1.0
kmf.services:name=offset-commit-service,type=offset-commit-service:offset-commit-service-success-rate=1.0
kmf.services:name=offset-commit-service,type=offset-commit-service:offset-commit-service-success-total=4.0
kmf.services:name=offset-commit-service,type=offset-commit-service:offset-commit-service-failure-rate=0.0
kmf.services:name=offset-commit-service,type=offset-commit-service:offset-commit-service-failure-total=0.0
 (com.linkedin.xinfra.monitor.services.DefaultMetricsReporterService)
```

### DIFF
--- a/src/main/java/com/linkedin/xinfra/monitor/services/metrics/OffsetCommitServiceMetrics.java
+++ b/src/main/java/com/linkedin/xinfra/monitor/services/metrics/OffsetCommitServiceMetrics.java
@@ -51,13 +51,13 @@ public class OffsetCommitServiceMetrics extends XinfraMonitorMetrics {
         new Total());
 
     _offsetCommitFailSensor = metrics.sensor(FAILURE_SENSOR_NAME);
+    /* NaN will persist as long as no record is submitted to the failure sensor.
+       we'll continue with NaN for now since we'd rather that the Sensor itself is a true and unaltered record of what values it recorded. */
     _offsetCommitFailSensor.add(new MetricName(FAILURE_RATE_METRIC, METRIC_GROUP_NAME,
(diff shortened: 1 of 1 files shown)


andrewchoi5:
@mgrubent  -- minor PR.

mgrubent:
Where are we trying to avoid NaN?

Is it possible to map NaN appropriately in that method?

andrewchoi5:
I don’t have any opinion on whether NaN is bad or good. I’m okay with persisting NaN. Just looks kinda strange when everything else is > 0. Also, maybe InGraph might be confused since it’s NaN. Whereas if we have 0 -> this means we have 0 percent failure rate.
Nonetheless, let’s continue with NaN for now; rather that the Sensor itself is a true and unaltered record of what values it recorded.

mgrubent:
Sounds good to me, thanks @andrewchoi5

andrewchoi5:
> Sounds good to me, thanks @andrewchoi5

Thanks, made one last change to note this behaviour in a code comment.

PR #296, opened 2020-07-16
path fix for docker/Dockerfile and bin/single-cluster-monitor.sh

talipkorkmaz:
After cloning the project, I found that `docker/Dockerfile` is not able to build due to the path error of kafka-monitor-start.sh. Also com/linkedin/kmf/apps/SingleClusterMonitor path was wrong in `bin/single-cluster-monitor.sh` file.

### DIFF
--- a/bin/single-cluster-monitor.sh
+++ b/bin/single-cluster-monitor.sh
@@ -9,4 +9,4 @@
 
 base_dir=$(dirname $0)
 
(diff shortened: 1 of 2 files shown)


talipkorkmaz:
what is needed to be done for passing greetings check?

andrewchoi5:
https://github.com/actions/first-interaction/issues/10
Seems like quite a few people are having issue with this greetings labeler. 
You can ignore it for now.

PR #317, opened 2020-10-06
 Kafka 2.4 + KIP-455 for Xinfra Monitor Multie Cluster Topic Management Service 

andrewchoi5:
`Kafka 2.4` + KIP-455 for Xinfra Monitor `Multi Cluster Topic Management Service` 

Total extends CumulativeSum : A non-sampled cumulative total maintained over all time.
 This is a non-sampled version of {@link WindowedSum}.
Total is deprecated since 2.4. Recommended to use {@link CumulativeSum} instead.

(New) CumulativeSum: A non-sampled cumulative total maintained over all time.
This is a non-sampled version of {@link WindowedSum}.

### DIFF
--- a/src/main/java/com/linkedin/xinfra/monitor/services/MultiClusterTopicManagementService.java
+++ b/src/main/java/com/linkedin/xinfra/monitor/services/MultiClusterTopicManagementService.java
@@ -34,6 +34,7 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
+import kafka.admin.AdminUtils;
 import kafka.admin.BrokerMetadata;
 import kafka.server.ConfigType;
 import kafka.zk.KafkaZkClient;
@@ -56,6 +57,8 @@ import org.apache.kafka.common.security.JaasUtils;
 import org.apache.kafka.common.utils.Time;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import scala.Option$;
+import scala.collection.JavaConverters;
 import scala.collection.Seq;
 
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/linkedin/xinfra/monitor/services/MultiClusterTopicManagementService.java
```diff
@@ -426,13 +431,12 @@ void maybeReassignPartitionAndElectLeader() throws Exception {
               "Configured replication factor {} is smaller than the current replication factor {} of the topic {} in cluster.",
               _replicationFactor, currentReplicationFactor, _topic);
 
-        if (expectedReplicationFactor > currentReplicationFactor && !zkClient
-            .reassignPartitionsInProgress()) {
+        if (expectedReplicationFactor > currentReplicationFactor
+            && Utils.ongoingPartitionReassignments(_adminClient).isEmpty()) {
```
gitlw, line 435: Can we check the ongoing partition reassignments only for this single topic?

gitlw: I mean passing the list of partitions, and call the API 
listPartitionReassignments(Set<TopicPartition> partitions)

andrewchoi5: We can. We wanted to replicate the previous behavior of `.reassignPartitionsInProgress()` which checks for all. 
Do we want to check only a set of topic partitions? @gitlw

andrewchoi5: I think you raise a very good point but in order to keep the exact same behaviour to be on a much safer side, we should probably keep the Utils.ongoingPartitionReassignments.isEmpty? — which is similar to  !zkClient.reassignPartitionsInProgress()
Let's check if getting all of the ongoing partition reassignments is costly when using the new API.

gitlw (approved):
LGTM

PR #336, opened 2021-02-04
Add two config properties to disable the add-topic-partition procedure and the topic-reassig-partition-and-leader-elect procedure respectively

Lincong:
Config props added:

1. `topic-management.topicAddPartitionEnabled`
2. `topic-management.topicReassignPartitionAndElectLeaderEnabled`

Both default to `true` for backward compatibility.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,6 @@ src/test/java/com/linkedin/xinfra/monitor/RandomTests.java
 config/andrew-choi.properties
 config/andrew-multi-cluster-monitor.properties
 
-
+kafka-monitor.iml
+kafka-monitor.ipr
+kafka-monitor.iws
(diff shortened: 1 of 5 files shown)


smccauliff:
Is this being separated because CC won't create additional partitions, but it will move leadership and otherwise do "balancing"?

Lincong:
> Is this being separated because CC won't create additional partitions, but it will move leadership and otherwise do "balancing"?

Yes. CC does not have the feature to add partition(s) for a topic currently. So we need to continue relying on KMF to add partitions for KMF topics.

This change gives us the option to stop KMF from rebalancing KMF topics and the rebalancing job will be done by CC (see [PR](https://github.com/linkedin/cruise-control/pull/1429))

smccauliff (approved):
LGTM

andrewchoi5 (approved):
I like that since CC doesn't add partitions for a topic, we're leveraging XM to add partitions for XM topics. This abstraction of XM task is good to have.

PR #338, opened 2021-02-06
Fix a bug so that the value of the config topic-management.preferredleader.election.check.interval.ms can be set to the MultiClusterTopicManagementService

Lincong:
In the `createMultiClusterTopicManagementServiceProps`, we need to move the value (if it exists) of the `PREFERRED_LEADER_ELECTION_CHECK_INTERVAL_MS_CONFIG` config to the `serviceProps`. Otherwise, the value does not take effect in the `MultiClusterTopicManagementService` class.


radai-rosenblatt (approved):

efeg (approved):
LGTM

PR #339, opened 2021-02-06
Refactor and clean up

Lincong:
1. Add logging to the MultiClusterTopicManagementService class
2. Add a timeout to some admin client requests
3. Rename some variables with non-inclusive language

### DIFF
--- a/src/main/java/com/linkedin/xinfra/monitor/services/MultiClusterTopicManagementService.java
+++ b/src/main/java/com/linkedin/xinfra/monitor/services/MultiClusterTopicManagementService.java
@@ -263,6 +263,7 @@ public class MultiClusterTopicManagementService implements Service {
 
     @SuppressWarnings("unchecked")
     TopicManagementHelper(Map<String, Object> props) throws Exception {
+
       TopicManagementServiceConfig config = new TopicManagementServiceConfig(props);
       AdminClientConfig adminClientConfig = new AdminClientConfig(props);
       String topicFactoryClassName = config.getString(TopicManagementServiceConfig.TOPIC_FACTORY_CLASS_CONFIG);
@@ -290,9 +291,17 @@ public class MultiClusterTopicManagementService implements Service {
               TopicManagementServiceConfig.TOPIC_FACTORY_PROPS_CONFIG) : new HashMap();
       _topicFactory =
           (TopicFactory) Class.forName(topicFactoryClassName).getConstructor(Map.class).newInstance(topicFactoryConfig);
-
       _adminClient = constructAdminClient(props);
       LOGGER.info("{} configs: {}", _adminClient.getClass().getSimpleName(), props);
+      logConfigurationValues();
+    }
+
+    private void logConfigurationValues() {
+      LOGGER.info("TopicManagementHelper for cluster with Zookeeper connect {} is configured with " +
+              "[topic={}, topicCreationEnabled={}, topicAddPartitionEnabled={}, " +
+              "topicReassignPartitionAndElectLeaderEnabled={}, minPartitionsToBrokersRatio={}, " +
+              "minPartitionNum={}]", _zkConnect, _topic, _topicCreationEnabled, _topicAddPartitionEnabled,
+              _topicReassignPartitionAndElectLeaderEnabled, _minPartitionsToBrokersRatio, _minPartitionNum);
     }
 
     @SuppressWarnings("unchecked")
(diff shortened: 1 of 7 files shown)


src/main/java/com/linkedin/xinfra/monitor/services/MultiClusterTopicManagementService.java
```diff
@@ -326,32 +338,37 @@ void maybeAddPartitions(int minPartitionNum) throws ExecutionException, Interrup
...
-          brokers.add(brokerMetadata);
-        }
+      List<TopicPartitionInfo> partitions = topicDescriptions.get(_requestTimeout.toMillis(), TimeUnit.MILLISECONDS).partitions();
 
-        if (!blackListedBrokers.isEmpty()) {
-          brokers.removeIf(broker -> blackListedBrokers.contains(broker.id()));
-        }
+      final int currPartitionNum = partitions.size();
+      if (currPartitionNum >= requiredMinPartitionNum) {
+        LOGGER.info("{} will not increase partition of the topic {} in the cluster. Current partition count {} and '" +
```
radai-rosenblatt, line 345: is tthis expected tto be the common case? if so - shuold this be  info?

Lincong: It is a common case. However, to troubleshoot the issue in which the KMF did not increase the partition number as expected, this `info` log line will be helpful since it provides the exact reason why KMF decided not to take any action to increase the partition number.

xiowu0: +1 on Radai's comment, since it is the common case, the info log may flush the log from 10+ monitored clusters periodically.

Lincong: OK, I changed it to the `debug` logging level

src/main/java/com/linkedin/xinfra/monitor/services/MultiClusterTopicManagementService.java
```diff
@@ -326,32 +338,37 @@ void maybeAddPartitions(int minPartitionNum) throws ExecutionException, Interrup
...
+      NewPartitions newPartitions = NewPartitions.increaseTo(requiredMinPartitionNum, newPartitionAssignments);
 
-        Map<String, NewPartitions> newPartitionsMap = new HashMap<>();
-        newPartitionsMap.put(_topic, newPartitions);
-        CreatePartitionsResult createPartitionsResult = _adminClient.createPartitions(newPartitionsMap);
-      }
+      Map<String, NewPartitions> newPartitionsMap = new HashMap<>();
+      newPartitionsMap.put(_topic, newPartitions);
+      CreatePartitionsResult createPartitionsResult = _adminClient.createPartitions(newPartitionsMap);
+      createPartitionsResult.all().wait(_requestTimeout.toMillis());
```
xiowu0, line 369: is it "object.wait(long)"?   use future.get ?

Lincong: Good idea

src/main/java/com/linkedin/xinfra/monitor/services/ProduceService.java
```diff
@@ -87,7 +87,11 @@ public ProduceService(Map<String, Object> props, String name) throws Exception {
     _partitioner = config.getConfiguredInstance(ProduceServiceConfig.PARTITIONER_CLASS_CONFIG, KMPartitioner.class);
     _threadsNum = config.getInt(ProduceServiceConfig.PRODUCE_THREAD_NUM_CONFIG);
     _topic = config.getString(ProduceServiceConfig.TOPIC_CONFIG);
-    _producerId = config.getString(ProduceServiceConfig.PRODUCER_ID_CONFIG);
+    // TODO: Some producer implementation does not allow multiple producers with the same producer ID. So, this is a
+    // temporary way to avoid producer ID collision. A better way to fix this issue is to let users provide producer ID
+    // prefix and KMF keeps track of the the producer service creation process and dynamically generate producer ID using
+    // the given prefix.
+    _producerId = String.format("%s-%s", _name, config.getString(ProduceServiceConfig.PRODUCER_ID_CONFIG));
```
xiowu0, line 94: should we honor the producer id set by config? 
this will confuse users if users intended to use a specific producer id and it was changed internally by the Produce service

Lincong: This change was intended to fix an existing issue. I will remove this change from this PR since it's not related to other changes in this PR. The producer ID issue will need another PR to fix. Thanks.

xiowu0 (approved):

Issue #325, opened 2020-10-28
InvalidReplicaAssignmentException: Duplicate replica assignment found


ppatierno:
I am running kafka-monitor (on Kubernetes) alongside a Kafka cluster with 3 brokers.
I have set the partitions to broker ratio to 1.0.
I see that kafka-monitor seems to work fine and it creates the xinfra-monitor-topic with 3 partitions:

```shell
Topic: xinfra-monitor-topic     PartitionCount: 3       ReplicationFactor: 1    Configs: retention.ms=3600000,message.format.version=2.6-IV0
        Topic: xinfra-monitor-topic     Partition: 0    Leader: 2       Replicas: 2     Isr: 2
        Topic: xinfra-monitor-topic     Partition: 1    Leader: 1       Replicas: 1     Isr: 1
        Topic: xinfra-monitor-topic     Partition: 2    Leader: 0       Replicas: 0     Isr: 0
```

but in the kafka-monitor log I see this exception more and more times

```shell
[2020-10-28 12:14:04,373] ERROR Exception occurred while retrieving the brokers count:  (com.linkedin.xinfra.monitor.services.ClusterTopicManipulationService)
java.util.concurrent.ExecutionException: org.apache.kafka.common.errors.InvalidReplicaAssignmentException: Duplicate replica assignment found: Map(2 -> ArrayBuffer(0, 1, 0), 1 -> ArrayBuffer(1, 0, 2), 0 -> ArrayBuffer(0, 2, 0))
    at org.apache.kafka.common.internals.KafkaFutureImpl.wrapAndThrow(KafkaFutureImpl.java:45) ~[kafka-clients-2.4.0.jar:?]
    at org.apache.kafka.common.internals.KafkaFutureImpl.access$000(KafkaFutureImpl.java:32) ~[kafka-clients-2.4.0.jar:?]
    at org.apache.kafka.common.internals.KafkaFutureImpl$SingleWaiter.await(KafkaFutureImpl.java:89) ~[kafka-clients-2.4.0.jar:?]
    at org.apache.kafka.common.internals.KafkaFutureImpl.get(KafkaFutureImpl.java:260) ~[kafka-clients-2.4.0.jar:?]
    at com.linkedin.xinfra.monitor.services.ClusterTopicManipulationService.createDeleteClusterTopic(ClusterTopicManipulationService.java:189) [kafka-monitor-2.5.3-SNAPSHOT.jar:2.5.3-SNAPSHOT]
[...]

PR #340, opened 2021-02-07
Fix replicasAssignment Hashmap

thkevin:
Hello,

This is a fix for the issue https://github.com/linkedin/kafka-monitor/issues/325. (Closed but not solved).
When creating a ClusterTopicManipulationService topic, the constructor uses a Hasmap containing the brokers id for each replica partition.
The problem is that has it is, the same broker id can be repeated.
With this fix, a simple shuffle of the brokers allows a random repartition of the brokers ids.
No need for the while loop.

Thanks

### DIFF
--- a/src/main/java/com/linkedin/xinfra/monitor/common/Utils.java
+++ b/src/main/java/com/linkedin/xinfra/monitor/common/Utils.java
@@ -19,10 +19,10 @@ import java.lang.management.ManagementFactory;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


zigarn (approved):

zigarn:
With the current code we have a simple random sampling *with replacement*, which have no sense for a replica assignment.

zigarn:
Any chance to see this reviewed and integrated? @andrewchoi5 @Lincong?

andrewchoi5 (commented):
lgtm, would also like to have @smccauliff give a +1.

andrewchoi5 (approved):

PR #344, opened 2021-03-23
Bump dependencies to fix security vulnerabilities

samzx:
Based on Snyk we have 3 vulneravilities that can be fixed. 2 dependencies can be bumped easily without issue. Kafka upgrade may be harder though. This PR bumps these two log4j and jackson databind.
image


andrewchoi5 (approved):
LGTM, @smccauliff can you also take a look?

PR #359, opened 2022-03-29
Use less ancient avro, maintain same wire format by @radai-rosenblatt

fluffywei:
Use less ancient avro, maintain same wire format by @radai-rosenblatt

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,12 +27,15 @@ allprojects {
 
   repositories {
     mavenCentral()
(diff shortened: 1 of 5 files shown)


mhratson:
Hi @fluffywei I noticed there's a dup #358, is it still needed?

fluffywei:
Thanks for pointing out @mhratson, #358 is no longer needed.

PR #362, opened 2022-05-06
Move config file format to YAML

MasonLegere:
- Updates the config to use YAML rather than a weird properties format
that is actually JSON but allows comments through the removal of lines
starting with `#`


MasonLegere:
@radai-rosenblatt Any chance I get some eyes on this?

PR #383, opened 2022-10-31
Use long for Connection timeout parameters for NetworkClient 

chromy96:
NetworkClient constructor expects Long parameters for conn.timeout: 
https://github.com/apache/kafka/blob/2.8/clients/src/main/java/org/apache/kafka/clients/NetworkClient.java#L142-L143


mhratson:
Thank you! addressed in #387 🙇

Issue #390, opened 2022-12-19
Ensure proper retry and backoff for newly created monitor topics


gitlw:
As shown in the conversations https://linkedin-randd.slack.com/archives/C04FMP0HB17/p1671222219329569,
if a new monitoring topic is just created in a cluster, the AdminClient.describeTopic API could result in UnknownTopicOrPartitionExceptions, which causes the whole process to crash. Below are the places that can trigger the exception (and there maybe more call sites)

https://github.com/linkedin/kafka-monitor/blob/7f99c095c2ceb2d09b0e490fa138a68fac849bba/src/main/java/com/linkedin/xinfra/monitor/services/MultiClusterTopicManagementService.java#L455

https://github.com/linkedin/kafka-monitor/blob/7f99c095c2ceb2d09b0e490fa138a68fac849bba/src/main/java/com/linkedin/xinfra/monitor/services/MultiClusterTopicManagementService.java#L338

We need to make sure that the logic calling the describeTopic API has appropriate retries and backoffs in case it's a topic that's just created.

PR #392, opened 2023-01-17
Prevent crashes by ensuring topics are created

mhratson:
addresses https://github.com/linkedin/kafka-monitor/issues/390

## Context

0. whenever a kafka-monitor starts monitoring a new cluster 
1. it has to create topics (if enabled)
2. monitoring process then describes the topics and does its job

## Problem
- due to possible delays between 1. and 2. 
- step 2 may crash, as due to propagation delay, metadata may not be yet available 
- this prevents kafka-monitor from starting as this may happen for multiple topics that may have to be created

## Solution
1. after creation query topic metadata
2. use retries and backoff to allow enough time for the metadata propagation

## Potential issues
1. slower start

### DIFF
--- a/src/main/java/com/linkedin/xinfra/monitor/services/MultiClusterTopicManagementService.java
+++ b/src/main/java/com/linkedin/xinfra/monitor/services/MultiClusterTopicManagementService.java
@@ -171,14 +171,15 @@ public class MultiClusterTopicManagementService implements Service {
   }
 
   private class TopicManagementRunnable implements Runnable {
+    private static final int MAX_TOPIC_FETCH_ATTEMPTS = 5;
 
     @Override
     public void run() {
       try {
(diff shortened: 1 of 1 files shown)


src/main/java/com/linkedin/xinfra/monitor/services/MultiClusterTopicManagementService.java
```diff
@@ -171,14 +171,15 @@ public void awaitShutdown(long timeout, TimeUnit unit) {
   }
 
   private class TopicManagementRunnable implements Runnable {
+    private static final int MAX_TOPIC_FETCH_ATTEMPTS = 5;
```
mhratson, line 174: Make configurable?

gitlw (approved):

mhratson:
will revisit the implementation later…

PR #393, opened 2023-03-22
ConsumeService: fix client closing causing `ConcurrentModificationExeption`

mhratson:
## Problem

- calling `_baseConsumer.close()`, outside of the thread consumer is running in, is invalid
- as docummented in _kafka consumer docs_[^1]

> The Kafka consumer is NOT thread-safe. All network I/O happens in the thread of the application * making the call. It is the responsibility of the user to ensure that multi-threaded access * is properly synchronized. Un-synchronized access will result in {@link ConcurrentModificationException}.

The exception thrown
```
2021/01/25 23:10:25.961 WARN [ConsumeService] [Thread-1] [kafka-monitoring] [] kac-lc/ConsumeService while trying to close consumer.
java.util.ConcurrentModificationException: KafkaConsumer is not safe for multi-threaded access. competing thread is kac-lc consume-service
        at com.linkedin.kafka.clients.utils.KafkaConsumerLock.lock(KafkaConsumerLock.java:31) ~[li-apache-kafka-clients-1.0.59.jar:?]
        at com.linkedin.kafka.clients.utils.CloseableLock.<init>(CloseableLock.java:18) ~[li-apache-kafka-clients-1.0.59.jar:?]
        at com.linkedin.kafka.clients.consumer.LiKafkaInstrumentedConsumerImpl.close(LiKafkaInstrumentedConsumerImpl.java:716) ~[li-apache-kafka-clients-1.0.59.jar:?]
```

## Solution

The recommended solution[^1] is
- to use `consumer.wakeup();` method
- but the method is not yet adopted by the `KMBaseConsumer` interface
- so for now `_baseConsumer.close()` is moved into the thread

[^1]:[KafkaConsumer.java](https://github.com/apache/kafka/blob/7d61d4505a16f09b85f5eb37adeff9c3534ec02c/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java#L467-L502)

## Testing Done

Increased `thread.join(5000)` timeout as this implementation is slower to stop due to not interrupting the consumer thread.

`- ./gradlew test`


src/main/java/com/linkedin/xinfra/monitor/services/ConsumeService.java
```diff
@@ -211,6 +214,10 @@ public void onComplete(Map<TopicPartition, OffsetAndMetadata> topicPartitionOffs
       }
     }
     /* end of consume() while loop */
+    LOG.info("{}/ConsumeService/Consumer closing.", _name);
+    _baseConsumer.close();
+    LOG.info("{}/ConsumeService/Consumer stopped.", _name);
+    Thread.sleep(1000);
```
lmr3796, line 220: `Thread.sleep` directly is a big bad smell in Java concurrent code. What's the purpose of this?

lmr3796 (changes requested):

src/main/java/com/linkedin/xinfra/monitor/services/ConsumeService.java
```diff
@@ -242,17 +248,18 @@ public synchronized void start() {
...
     }
   }
 
   @Override
   public void awaitShutdown(long timeout, TimeUnit unit) {
+    LOG.info("{}/ConsumeService shutdown awaiting…", _name);
+    try {
+      _consumeThread.join(unit.toMillis(timeout));
+    } catch (InterruptedException e) {
+      LOG.error(_name + "/ConsumeService interrupted", e);
```
lmr3796, line 261: This silently swallows the error here.  Is this `awaitShutdown` only invoked when the whole service is shutdown?

If not, we should either
1. Let the caller handle this or
2. retry and throw a runtime exception

lmr3796 (changes requested):

src/main/java/com/linkedin/xinfra/monitor/services/ConsumeService.java
mhratson: I think it's ok, because: 
1. exception is logged
2. that's [the way it's handled](https://github.com/linkedin/kafka-monitor/search?q=awaitShutdown+InterruptedException) in this project.

We could implement your suggestions but for consistency IMO, it should be:
1. kept the same way or
2. refactored across the project

WDYT?

lmr3796 (approved):

